PR #77, opened 2018-04-16
Remove empty string check as metric name can be blank.

titusjaka:
Sometimes metric name can be blank, so empty string check is redundant.

Moira doesn't have to throw error when metric name is empty.

firefox_2018-04-16_17-08-14


titusjaka:
Checked on Staging. Everything is 👌

PR #87, opened 2018-05-25
Fix bug when event was sent after maintenance even if state didn't change

titusjaka:


titusjaka:
@borovskyav вроде всё сделал. Жду ревью :)

PR #99, opened 2018-07-27
Add pattern matching workers

errx:
We have encountered a performance issues with moira-filter (slow metrics ingestion which resulted in false-positive NODATA alerts)

The problem is that we have a large number of incoming metrics from a single host (carbon-c-relay in our case).
Current implementation assigns single goroutine per connections for a relatively slow process of pattern matching.

By introducing separate pool of pattern matching workers we completely solved our problem with slow ingestion

### DIFF
--- a/cmd/filter/main.go
+++ b/cmd/filter/main.go
@@ -58,7 +58,7 @@ func main() {
 	}
 
 	if config.Filter.MaxParallelChecks == 0 {
-		fmt.Fprint(os.Stderr, "MaxParallelChecks does not configured, filter does not started")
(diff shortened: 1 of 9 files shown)


cmd/filter/main.go
```diff
@@ -57,6 +57,10 @@ func main() {
 		os.Exit(1)
 	}
 
+	if config.Filter.MaxParallelChecks == 0 {
+		fmt.Fprint(os.Stderr, "MaxParallelChecks does not configured, filter does not started")
```
titusjaka: Правильнее написать:
> MaxParallelChecks *is* not configured, filter does not *start*

PR #105, opened 2018-08-07
Fix accident bug brought from #94

titusjaka:


titusjaka:
@kamaev просто нажми approve

image

kamaev:
house-working

PR #108, opened 2018-08-13
Fix default MaxParallelMatches option. If not set or equal 0, set it to number of CPU

titusjaka:
related with #99


titusjaka:
@borovskyav вливаю?

PR #115, opened 2018-08-27
remove trigger events key on trigger removal

kamaev:


kamaev:
Switched to correct branch: https://github.com/moira-alert/moira/pull/116

PR #279, opened 2019-03-02
Limit message size in Slack sender

### LINKED ISSUE #255: Limit message size in Slack sender
Slack has a limit on message size, it's 40000 symbols (see https://api.slack.com/changelog/2018-04-truncating-really-long-messages). Our Slack sender currently ignores this limit, which leads to truncated and malformed messages. We need to implement this limit in our Slack sender in a similar fashion to Telegram message size limit (see https://github.com/moira-alert/moira/blob/8fb128733d66a2b2a4121c5cf147202dd9fecead/senders/telegram/init.go#L15).

Slack sender is here: https://github.com/moira-alert/moira/blob/master/senders/slack/slack.go

iurimateus:
Fix #255
Solution is based on Telegram's sender.

Even though the original issue mentioned a 40k character limit, the current limit set by Slack seems to be at 4k.

### DIFF
--- a/senders/slack/slack.go
+++ b/senders/slack/slack.go
@@ -19,7 +19,7 @@ const (
 	exceptionEmoji = ":moira-state-exception:"
 	testEmoji      = ":moira-state-test:"
 
-	messageMaxCharacters          = 4000
+	messageMaxCharacters          = 40000
 	additionalInfoCharactersCount = 400
 )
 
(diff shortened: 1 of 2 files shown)


iurimateus:
> Hi! Thanks for your work. I see two posts ([one](https://api.slack.com/docs/message-formatting), [two](https://api.slack.com/changelog/2018-04-truncating-really-long-messages)) about 40000 symbols message limit.

I think that's indeed the API limit. However my client was breaking at 4000 in multiple messages. In any case, it should be long enough.

borovskyav:
In Slack case we has one important difference unlike telegram sender: in slack sender we render trigger description. It can be unlimited size. So in slack cases we care about message size and we really need this 40000 symbols.

PR #290, opened 2019-03-05
Add smtp_hello parameter in the mail sender

### LINKED ISSUE #289: Add smtp_hello parameter
Hello.
In some cases, it is necessary to explicitly set smtp_hello in the mail sending settings, now there is no such possibility

mikaelmello:
Closes #289 

The default LocalName (the hostname sent to the SMTP server with the HELO command) in the Dialer struct is "localhost". Therefore when the `smtp_hello` parameter is not provided, we set `Sender.SMTPhello` to `localhost` in order to keep using the default

### DIFF
--- a/senders/mail/mail.go
+++ b/senders/mail/mail.go
@@ -18,6 +18,7 @@ import (
 // Sender implements moira sender interface via pushover
 type Sender struct {
 	From           string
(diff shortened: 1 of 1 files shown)


mikaelmello:
I'll redo it based on `develop`

PR #291, opened 2019-03-06
Add smtp_hello parameter on mail sender

### LINKED ISSUE #289: Add smtp_hello parameter

mikaelmello:
Closes #289

The default LocalName (the hostname sent to the SMTP server with the HELO command) in the Dialer struct is "localhost". Therefore if the `smtp_hello` parameter is not provided, `sender.SMTPHello` is set to `""`, which then falls back to the default LocalName: https://github.com/go-gomail/gomail/blob/81ebce5c23dfd25c6c67194b37d3dd3f338c98b1/smtp.go#L75-L79

### DIFF
--- a/senders/mail/mail.go
+++ b/senders/mail/mail.go
@@ -15,6 +15,7 @@ import (
 // Sender implements moira sender interface via pushover
 type Sender struct {
 	From           string
+	SMTPHello			 string
 	SMTPHost       string
 	SMTPPort       int64
 	FrontURI       string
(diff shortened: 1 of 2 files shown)


borovskyav:
Hi! Thanks for your work, you used smpt_hello while send message but we has tryDial on init Sender and wanna send smpt_hello in this case too.

mikaelmello:
@borovskyav, done

PR #293, opened 2019-03-08
Limit the URL and title in pushover sender

### LINKED ISSUE #250: Limit message size in Pushover sender
Pushover API has a limit on length of various fields in message. We need to implement these restrictions in our Pushover sender.

API docs: https://pushover.net/api#messages
Library constants: https://github.com/gregdel/pushover/blob/master/pushover.go#L46
Pushover sender: https://github.com/moira-alert/moira/blob/master/senders/pushover/pushover.go

We need to update gregdel/pushover library (see fix below) and then:
1. Restrict Title length.
2. Restrict message length (probably already done).
3. Restrict URL length.

aswinmprabhu:
closes #250

### DIFF
--- a/senders/pushover/pushover.go
+++ b/senders/pushover/pushover.go
@@ -12,6 +12,7 @@ import (
 
 const printEventsCount int = 5
 const titleLimit = 250
+const urlLimit = 512
 
 // Sender implements moira sender interface via pushover
 type Sender struct {
@@ -54,12 +55,14 @@ func (sender *Sender) makePushoverMessage(events moira.NotificationEvents, conta
 		Message:   sender.buildMessage(events, throttled),
 		Title:     sender.buildTitle(events, trigger),
 		Priority:  sender.getMessagePriority(events),
-		URL:       trigger.GetTriggerURI(sender.frontURI),
 		Retry:     5 * time.Minute,
 		Expire:    time.Hour,
 		Timestamp: events[len(events)-1].Timestamp,
 	}
-
+	url := trigger.GetTriggerURI(sender.frontURI)
+	if len(url) < urlLimit {
+		pushoverMessage.URL = url
+	}
 	if len(plot) > 0 {
 		reader := bytes.NewReader(plot)
 		pushoverMessage.AddAttachment(reader)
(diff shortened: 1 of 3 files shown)


senders/pushover/pushover.go
```diff
@@ -91,7 +95,17 @@ func (sender *Sender) buildMessage(events moira.NotificationEvents, throttled bo
 }
 
 func (sender *Sender) buildTitle(events moira.NotificationEvents, trigger moira.TriggerData) string {
-	return fmt.Sprintf("%s %s %s (%d)", events.GetSubjectState(), trigger.Name, trigger.GetTags(), len(events))
+	title := fmt.Sprintf("%s %s %s (%d)", events.GetSubjectState(), trigger.Name, trigger.GetTags(), len(events))
+	tags := 1
+	for len(title) > titleLimit-30 {
```
borovskyav: We should use ```len([]rune(title))``` instead ```len(title)``` because ```len(title)``` returns size in bytes but we need to check symbols len

senders/pushover/pushover.go
```diff
@@ -91,7 +95,17 @@ func (sender *Sender) buildMessage(events moira.NotificationEvents, throttled bo
...
+		var tagBuffer bytes.Buffer
+		for i := 0; i < len(trigger.Tags)-tags; i++ {
+			tagBuffer.WriteString(fmt.Sprintf("[%s]", trigger.Tags[i]))
+		}
+		title = fmt.Sprintf("%s %s %s and %d more tags (%d)", events.GetSubjectState(), trigger.Name, tagBuffer.String(), tags, len(events))
```
borovskyav: Let's change this logic:
In your solution you reserve 30 symbols for ```and %d more tags``` notation but in typical cases tag len is short (examples: [FreeSpace] or [Datacenter]). Lets cut off tags silently

senders/pushover/pushover.go
borovskyav: Oh, you just should not use ```titleLimit-30```. Only ```titleLimit``` because in your case you build title in every cycle, but in other or sender's we reverse symbos for complete message.

senders/pushover/pushover.go
borovskyav: Oh, let's use ```...``` instead of ```and %d more tags```

aswinmprabhu:
@borovskyav, Can you please review the changes I made on your request ?

PR #294, opened 2019-03-09
Second step of tags support

Pliner:


filter/series_by_tag_test.go
```diff
@@ -0,0 +1,72 @@
...
+	Convey("Given patterns with tagspecs, should build index and match patterns", t, func() {
+		tagSpecsByPattern := map[string][]TagSpec{
+			"name=cpu1":        {{"name", EqualOperator, "cpu1"}},
+			"name!=cpu1":       {{"name", NotEqualOperator, "cpu1"}},
+			"name~=cpu":        {{"name", MatchOperator, "cpu"}},
+			"name!~=cpu":       {{"name", NotMatchOperator, "cpu"}},
+			"dc=ru1":           {{"dc", EqualOperator, "ru1"}},
+			"dc!=ru1":          {{"dc", NotEqualOperator, "ru1"}},
+			"dc~=ru":           {{"dc", MatchOperator, "ru"}},
+			"dc!~=ru":          {{"dc", NotMatchOperator, "ru"}},
```
borovskyav: Я добавил вот такие два спека:
```
"name=cpu1&&dc=ru1": {{"name", EqualOperator, "cpu1"}, {"dc", EqualOperator, "ru1"}},
"name=cpu1&&dc=ru2": {{"name", EqualOperator, "cpu1"}, {"dc", EqualOperator, "ru2"}},
```
Ожидал что кейс 
```
{"cpu1", map[string]string{}, []string{"name=cpu1", "name~=cpu"}},
```
пройдет. Или я что-то неправильно ожидал?

biancalana: would be easier if default language was english...

Pliner: @borovskyav I fixed your comment. Actually, the filtering by multiple conditions on the same tag didn't work. Now it works.

filter/series_by_tag_test.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	Convey("Given valid seriesByTag patterns, should return parsed tag specs", t, func() {
+		validSeriesByTagCases := []ValidSeriesByTagCase{
+			{"seriesByTag(\"a=b\")", []TagSpec{{"a", EqualOperator, "b"}}},
+			{"seriesByTag(\"a!=b\")", []TagSpec{{"a", NotEqualOperator, "b"}}},
+			{"seriesByTag(\"a=~b\")", []TagSpec{{"a", MatchOperator, "b"}}},
+			{"seriesByTag(\"a!=~b\")", []TagSpec{{"a", NotMatchOperator, "b"}}},
+			{"seriesByTag(\"a=\")", []TagSpec{{"a", EqualOperator, ""}}},
+			{"seriesByTag(\"a=b\",\"a=c\")", []TagSpec{{"a", EqualOperator, "c"}}},
```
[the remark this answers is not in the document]
Pliner: >{"seriesByTag(\"a=b\",\"b!=c\",\"m=~p\")", []TagSpec{{"a", EqualOperator, "b"}, {"b", NotEqualOperator, "c"}, {"m", MatchOperator, "p"}}}

Now we can parse multiple filter conditions. Before the fix parsing was totally broken.

Pliner:
> Кстати, в доке еще вот такое есть:
> 
> ```
> Any tag spec that matches an empty value is considered to match series that don’t have that tag.
> 
> At least one tag spec must require a non-empty value.
> ```
> Чет я не понял, будет ли у нас такое работать.

It doesn't work. Will fix it asap.

biancalana:
I've found some issues here, the filter was returning metrics that matched any specified tag the right thing is to match all. Also changed a the regex that match SeriesByTag so it also accepts single quotes on parameters. I have sent that throught #330

biancalana:
Now it’s working like a charm! The last thing I’m going to test is groupByTags

PR #330, opened 2019-03-31
Feature/tags series by tag

biancalana:
Hi !

Testing that feature, I found two issues and fix then:

- Only accepting SeriesByTag parameters when using double quotes
- Filters was matching any tag of metric that way, the right thing to do is to match all tags specified

### DIFF
--- a/filter/series_by_tag.go
+++ b/filter/series_by_tag.go
@@ -130,6 +130,10 @@ func (index *SeriesByTagPatternIndex) MatchPatterns(name string, labels map[stri
 		}
 	}
 
+	if len(index.filtersByTag) != len(matchedPatterns) {
+		return ([]string{})
+	}
+
 	return moira.GetStringListsUnion(matchedPatterns...)
 }
 
--- a/filter/series_by_tag.go
+++ b/filter/series_by_tag.go
@@ -7,7 +7,7 @@ import (
 	"github.com/moira-alert/moira"
 )
 
-var tagSpecRegex = regexp.MustCompile(`^"([^,!=]+)\s*(!?=~?)\s*([^,]*)"`)
+var tagSpecRegex = regexp.MustCompile(`^["']{1}([^,!=]+)\s*(!?=~?)\s*([^,]*)["']{1}`)
 var tagSpecDelimiterRegex = regexp.MustCompile(`^\s*,\s*`)
 var seriesByTagRegex = regexp.MustCompile(`^seriesByTag\(([^)]+)\)$`)
 


filter/series_by_tag.go
```diff
@@ -130,6 +130,10 @@ func (index *SeriesByTagPatternIndex) MatchPatterns(name string, labels map[stri
 		}
 	}
 
+	if len(index.filtersByTag) != len(matchedPatterns) {
```
Pliner: Could you explain why we need this?

biancalana: In my testing, sending two metrics:
```
metricOne.test;region=us-east-2;az=a;service=foo
metricTwo.test;region=us-east-2;az=a;service=foo
```

Having the following trigger target:
```
seriesByTag("name=metricOne.test", 'region=us-east-2', "az=a", "service=foo")
```

Make both metrics pass the filter and arrive checker, when only the first should.

My conclusion is that 
```
func (index *SeriesByTagPatternIndex) MatchPatterns
```
returns metrics that matched any of the given tags. My understanding is that it should only return metrics that match all given tags, because of that a check that compare the number off filtertags and matched patterns.

biancalana: After additional testing the main thing here (before this pr) is that even when metric doesn't match name it keep going and if there's return specs that match any tag.
Other way to fix is to simply return if name doesn't match.

Besides that after advancing with test I realized that tag with empty value (that should match series that does not have the tag) also does not work, series keep matching...

Pliner: Yep, I will fix it.

Pliner: Actually, it seems your fix is not correct in general, but I will try to rework it, so many thanks for reporting the issue.

Pliner: BTW Here is a test https://github.com/moira-alert/moira/pull/294/commits/469a511fd8ac48a70b9602c8eecbb0edbc697b20

biancalana: agreed, looking the code for more time I understand kkkk

Pliner: Here is a rework of the `SeriesByTagPatternIndex` https://github.com/moira-alert/moira/pull/294/commits/8e60157d3c2ed2d6640684a326514ecfd6558ee4. It is simple enough but I hope it covers most of the cases.

Pliner: Also we added single quoted tag spec https://github.com/moira-alert/moira/pull/294/commits/6af3cbd738dd3278987cbddc8b416623f360ef70

Pliner:
@biancalana Thank you. If you have any suggestions feel free to open new PR :wink:

Issue #353, opened 2019-04-16
Markdown descriptions in senders


beevee:
Web UI supports Markdown descriptions since 2.5 (see #285), but Slack and other senders do not. We need to send descriptions as Markdown in all sender types that support it. And also maybe we need to convert Markdown for senders that support HTML or other types of markup languages.

aswinmprabhu:
Hey. I can help. Slack supports basic MD I guess.
https://api.slack.com/docs/message-formatting#message_formatting

aswinmprabhu:
Senders like pushover and slack seem to support basic HTML but not all of it. For example, the web UI supports MD headers which are converted to `h1` tags but these are not supported by pushover and slack.
https://pushover.net/api#html
What do we do in such cases?

beevee:
We should:

1. Make sure all existing senders use Markdown properly, or at least strip formatting if they don't. For example, we can use full Markdown in Slack, we can convert to full unrestricted HTML for email, and to subset of HTML for Pushover, etc.
2. Make sure all existing senders use description to some extent, depending on message size limit. For example, Slack can use full description, Telegram can use first N characters of description, etc. SMS sender should probably use 0 characters of description :-)

For example, we can build a message for Slack (4000 characters limit) this way:
1. Render title, tags and all other short fields. Let's say this takes 100 characters, and we have 3900 left.
2. Render description, limited to some amount of characters that is less than total amount left. Let's say 3000. So even if we have a long description, we still have some characters left (900 in this case).
3. Render metric events with all characters left. If we don't have enough characters left, we can use "and N more events", like we do now.

aswinmprabhu:
Slack's MD seems to a bit different.
The normal MD syntax supported by the web UI uses 2 * around the text for bold but slack uses just one * around the text.
Another point is that telegram and slack don't support MD headers that start with #. What should be done in these cases?

beevee:
Try to convert to appropriate syntax where possible, and strip all unsupported syntax, leaving just plain text.

beevee:
Message and description size fixed for Slack. Markdown formatting added to Slack. Description added for Email. #403

beevee:
Other senders, like Telegram and Pushover, are not ready yet.

PR #370, opened 2019-05-16
Add MD support for all senders (#353)

aswinmprabhu:


senders/slack/slack.go
```diff
@@ -89,15 +90,32 @@ func (sender *Sender) buildMessage(events moira.NotificationEvents, trigger moir
...
+	messageCharsCount := len([]rune(message.String()))
+
 	if trigger.Desc != "" {
 		message.WriteString("\n")
-		message.WriteString(trigger.Desc)
+		charsRequiredForEvents := 9000
+		charsAvailableForDesc := messageMaxCharacters - messageCharsCount - charsRequiredForEvents - 1
+
+		// Replace **bold text** with *bold text* that slack supports
+		re := regexp.MustCompile(`(?m)\*\*(?P<boldtext>[\w\s]+)\*\*`)
```
kamaev: Once we got our application started:
- how many times sender.SendEvents() will be executed?
- and how many times sender.Init() will be executed?
:)

kamaev:
Guys, look at this:
https://github.com/karriereat/blackfriday-slack
maybe this is some kind of preprocessing we are looking for?

aswinmprabhu:
I had tried using the blackfriday-slack package. It had a dependency issue. It seems to depend on github.com/russross/blackfriday/v2 while the README for the library says the import path is gopkg.in/russross/blackfriday.v2

Also since the processing required was minor, I thought a custom solution would be better than adding another dependency. I can give blackfriday-slack another look if need be.

I'll definitely add the tests!

beevee:
Please, read again my comment from the original issue: https://github.com/moira-alert/moira/issues/353#issuecomment-491828385

I don't think that you've paid enough attention to message length limits. For example, you did just append description to Pushover message without limiting message length in any way. Typical description can be several thousand characters long. Pushover message length is limited to 1024 characters (https://pushover.net/api#limits).

Please, review and thoroughly test your new senders. Try longer descriptions, more events. Attach screenshots. Invest more time in this task, because alert legibility is very important for our users.

aswinmprabhu:
Sorry, I missed message limiting in the case of pushover. Although I had taken care of it for the others. Anyway, I added message limiting to pushover and tests with long description to all senders.

senders/pushover/pushover.go
```diff
@@ -71,21 +75,34 @@ func (sender *Sender) makePushoverMessage(events moira.NotificationEvents, conta
...
-			break
-		}
-		message.WriteString(fmt.Sprintf("%s: %s = %s (%s to %s)", event.FormatTimestamp(sender.location), event.Metric, event.GetMetricValue(), event.OldState, event.State))
+	desc := trigger.Desc
+	if len(desc) > descLimit {
+		// trim description to fit 512 chars while leaving space for html tags
+		desc = desc[:descLimit-100] + "...\n"
+	}
+	htmlDesc := blackfriday.Run([]byte(desc))
+	htmlDescWithbr := strings.Replace(string(htmlDesc), "\n", "\n<br/>", -1)
```
beevee: Why do we need both `\n` and `<br/>` in resulting text? Maybe just `<br/>` is enough?

senders/pushover/pushover.go
```diff
@@ -71,21 +75,34 @@ func (sender *Sender) makePushoverMessage(events moira.NotificationEvents, conta
...
+func (sender *Sender) buildMessage(events moira.NotificationEvents, throttled bool, trigger moira.TriggerData) string {
 	var message bytes.Buffer
-	for i, event := range events {
-		if i > printEventsCount-1 {
-			break
-		}
-		message.WriteString(fmt.Sprintf("%s: %s = %s (%s to %s)", event.FormatTimestamp(sender.location), event.Metric, event.GetMetricValue(), event.OldState, event.State))
+	desc := trigger.Desc
+	if len(desc) > descLimit {
+		// trim description to fit 512 chars while leaving space for html tags
```
beevee: It's a bad idea to write 512 in this comment. Constant values can change, and it will lead to confusion if somebody forgets to edit this comment in the future.

senders/pushover/pushover.go
```diff
@@ -71,21 +75,34 @@ func (sender *Sender) makePushoverMessage(events moira.NotificationEvents, conta
 	return pushoverMessage
 }
 
-func (sender *Sender) buildMessage(events moira.NotificationEvents, throttled bool) string {
+func (sender *Sender) buildMessage(events moira.NotificationEvents, throttled bool, trigger moira.TriggerData) string {
 	var message bytes.Buffer
```
beevee: Please, migrate this code to `strings.Builder`: https://medium.com/@felipedutratine/string-concatenation-in-golang-since-1-10-bytes-buffer-vs-strings-builder-2b3081848c45

senders/pushover/pushover.go
```diff
@@ -71,21 +75,34 @@ func (sender *Sender) makePushoverMessage(events moira.NotificationEvents, conta
...
+	for _, event := range events {
+		line := fmt.Sprintf("%s: %s = %s (%s to %s)", event.FormatTimestamp(sender.location), event.Metric, event.GetMetricValue(), event.OldState, event.State)
 		if len(moira.UseString(event.Message)) > 0 {
-			message.WriteString(fmt.Sprintf(". %s\n", moira.UseString(event.Message)))
+			line += fmt.Sprintf(". %s\n", moira.UseString(event.Message))
 		} else {
-			message.WriteString("\n")
+			line += "\n"
+		}
+		if len([]rune(message.String()+line)) > msgLimit-30 {
```
beevee: This code seems inefficient, because it copies `message.String()+line` to a new location each time, just to count runes. Rewrite this with a length counter, maybe?

senders/pushover/pushover.go
```diff
@@ -71,21 +75,34 @@ func (sender *Sender) makePushoverMessage(events moira.NotificationEvents, conta
...
+	htmlDesc := blackfriday.Run([]byte(desc))
+	htmlDescWithbr := strings.Replace(string(htmlDesc), "\n", "\n<br/>", -1)
+	message.WriteString(htmlDescWithbr)
+	msgLimitReached := false
+	eventsPrinted := 0
+	for _, event := range events {
+		line := fmt.Sprintf("%s: %s = %s (%s to %s)", event.FormatTimestamp(sender.location), event.Metric, event.GetMetricValue(), event.OldState, event.State)
 		if len(moira.UseString(event.Message)) > 0 {
-			message.WriteString(fmt.Sprintf(". %s\n", moira.UseString(event.Message)))
+			line += fmt.Sprintf(". %s\n", moira.UseString(event.Message))
```
beevee: Don't you also need to replace this `\n` with `<br/>`?

senders/pushover/pushover.go
```diff
@@ -71,21 +75,34 @@ func (sender *Sender) makePushoverMessage(events moira.NotificationEvents, conta
...
 		} else {
-			message.WriteString("\n")
+			line += "\n"
```
beevee: br?

senders/pushover/pushover.go
```diff
@@ -71,21 +75,34 @@ func (sender *Sender) makePushoverMessage(events moira.NotificationEvents, conta
...
+			msgLimitReached = true
+			break
 		}
+		message.WriteString(line)
+		eventsPrinted++
 	}
-	if len(events) > printEventsCount {
-		message.WriteString(fmt.Sprintf("\n...and %d more events.", len(events)-printEventsCount))
+	if msgLimitReached {
+		message.WriteString(fmt.Sprintf("\n...and %d more events.", len(events)-eventsPrinted))
```
beevee: br?

senders/slack/slack.go
```diff
@@ -21,6 +22,12 @@ const (
 
 	messageMaxCharacters          = 40000
```
beevee: Slack messages are actually limited to 4000 chars: https://api.slack.com/docs/rate-limits

Actually, it's 16 kilobytes for the entire JSON, including metadata. But 4000 chars is safe for any UTF-8 content.

beevee:
Generally, I'm not satisfied with this algorithm. In short, that's what it does:

1. Take some arbitrary limitation for description size. Maybe it's 512 chars out of 1024 total, like in the Pushover sender.
2. Try to fit description into this limitation. Ellipsize if it doesn't fit.
3. Use remaining chars for events.
(I'm simplifying things and omitting other metadata)

Consider three cases:

Case 1: description is 400 chars, events are 400 chars. All good, total message size 800.
Case 2: description is 400 chars, events are 1000 chars. All good, events get shortened, so total message size is 1024.
Case 3: description is 600 chars, events are 400 chars. Description gets shortened to 512, total message size is 912 chars. We could do 1024 and fit the entire description! Why did we shorten it?

@aswinmprabhu I think, the way forward is to carefully write some tests for corner cases first. Try to fit long descriptions and large event counts. Try out on real Slack and Pushover notifications. Think about best ways to fit the content. Do not consider current implementations to be the best ones (seems like you just took the code from Telegram sender here). Current implementations need to be improved.

beevee:
(Yes, I know that you've implemented the algorithm I proposed in #353. But I see this flaw now very clearly)

beevee:
Also, there is a bug in almost every sender: `Please, fix your system or tune this trigger to generate less events.` line is not counted towards size limit. Problem is, this line appears in throttled messages, which are usually the largest ones, and more likely not to fit in character limit.

senders/pushover/pushover.go
aswinmprabhu: <br> is required for email but not for pushover. Removed!

senders/pushover/pushover.go
aswinmprabhu: As of Go 1.11, the compiler optimizes this pattern to behave the same as utf8.RuneCountInString(string)
https://github.com/golang/go/issues/24923

senders/pushover/pushover.go
aswinmprabhu: As of Go 1.11, the compiler optimizes this pattern to behave like `utf8.RuneCountInString(string)` https://github.com/golang/go/issues/24923

senders/slack/slack.go
aswinmprabhu: Will fix this once the algorithm for message building is finalized.

aswinmprabhu:
I was not aware that you are unhappy with the implementation of the existing senders. I will rewrite the message building algorithms from scratch then. I wrote an algorithm that solves your concerns for the pushover sender. This can be adopted for all senders if it is acceptable. I am not sure how to test it though, without rewriting the algorithm logic in the test code. Manually constructing the expected string by hand is one option.

beevee:
Just a minor note: there is not even one empty line in your code separating logically different blocks of code. In some places even functions are not separated. In my opinion, it makes code harder to read.

Unfortunately, I can't offer you any solid advice or guidelines on where to put these empty lines. But I instinctively know that 0 is too few :-)

senders/pushover/pushover.go
```diff
@@ -71,29 +74,85 @@ func (sender *Sender) makePushoverMessage(events moira.NotificationEvents, conta
...
+		eventsString = sender.buildEventsString(events, charsForEvents, throttled)
+	}
+	message.WriteString(htmlDesc)
+	message.WriteString(eventsString)
+	return message.String()
+}
+func (sender *Sender) buildEventsString(events moira.NotificationEvents, charsForEvents int, throttled bool) string {
+	charsForThrottleMsg := 0
+	if throttled {
+		charsForThrottleMsg = 70
```
beevee: It's bad to hardcode `70` here. Somebody will change the throttling text and forget to update this value. Better make a constant with the text and use its length.

senders/pushover/pushover.go
```diff
@@ -71,29 +74,85 @@ func (sender *Sender) makePushoverMessage(events moira.NotificationEvents, conta
...
+	for _, event := range events {
+		line := fmt.Sprintf("%s: %s = %s (%s to %s)", event.FormatTimestamp(sender.location), event.Metric, event.GetMetricValue(), event.OldState, event.State)
 		if len(moira.UseString(event.Message)) > 0 {
-			message.WriteString(fmt.Sprintf(". %s\n", moira.UseString(event.Message)))
+			line += fmt.Sprintf(". %s\n", moira.UseString(event.Message))
 		} else {
-			message.WriteString("\n")
+			line += "\n"
 		}
+		if len([]rune(eventsString+line)) > charsLeftForEvents {
```
beevee: I'm returning again to this line. Yes, as you've correctly mentioned in a previous discussion, `len([]rune(somestring))` works fast. But the problem is with `+`. When you do `eventsString+line`, Go allocates another string of length `len(eventsString)+len(line)` and _copies_ contents of two variables there. A whole lot of copying just to count the sum of lengths.

`len([]rune(eventsString)) + len([]rune(line))` should be much faster, since this code does not copy anything.

senders/pushover/pushover.go
```diff
@@ -71,29 +74,85 @@ func (sender *Sender) makePushoverMessage(events moira.NotificationEvents, conta
...
+	} else if htmlDescLen > msgLimit/2 {
+		// Trim the desc to the chars left after using the whole events string
+		charsForDesc := msgLimit - eventsStringLen
+		desc = desc[:charsForDesc-charsForHTMLTags-10] + "...\n"
+		htmlDesc = string(blackfriday.Run([]byte(desc)))
+	} else if eventsStringLen > msgLimit/2 {
+		// Trim the events string to the chars left after using the whole desc
+		charsForEvents := msgLimit/2 - htmlDescLen
+		eventsString = sender.buildEventsString(events, charsForEvents, throttled)
+	}
```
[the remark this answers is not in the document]
aswinmprabhu: I don't think a case will fail to match all if conditions. But you are right. Better to be safe!

senders/pushover/pushover.go
```diff
@@ -71,29 +74,85 @@ func (sender *Sender) makePushoverMessage(events moira.NotificationEvents, conta
...
-			break
+func (sender *Sender) buildMessage(events moira.NotificationEvents, throttled bool, trigger moira.TriggerData) string {
+	var message strings.Builder
+	desc := trigger.Desc
+	htmlDesc := string(blackfriday.Run([]byte(desc)))
+	descLen := len([]rune(desc))
+	htmlDescLen := len([]rune(htmlDesc))
+	charsForHTMLTags := htmlDescLen - descLen
+	var eventsString string
+	for _, event := range events {
```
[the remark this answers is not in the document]
aswinmprabhu: The behavior of buildEventsString() is a bit different as it limits the string. I modified it to build the whole string if charsForEvents is negative (similar to https://golang.org/pkg/strings/#SplitN).

senders/pushover/pushover.go
```diff
@@ -71,27 +74,94 @@ func (sender *Sender) makePushoverMessage(events moira.NotificationEvents, conta
...
+		charsForEvents := msgLimit - htmlDescLen
+		eventsString = sender.buildEventsString(events, charsForEvents, throttled)
+
+	} else {
+		// Trim both desc and events to half the message size each
+		desc = desc[:msgLimit/2-charsForHTMLTags-10] + "...\n"
+		htmlDesc = string(blackfriday.Run([]byte(desc)))
+		eventsString = sender.buildEventsString(events, msgLimit/2, throttled)
+
+	}
```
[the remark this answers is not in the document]
aswinmprabhu: I merged the lines 88...93 as a wrapping if block, so that it doesn't become an empty block. So from your example, if (1) is not true then consider the cases (2-4).

borovskyav:
Waiting for the generalization of logic, which I quoted in [comment](https://github.com/moira-alert/moira/pull/387#issuecomment-511340274).

aswinmprabhu:
The import path github.com/russross/blackfriday/v2 doesn't work with go get or govendor.
image
The official import path for v2 mentioned in the README involves gopkg. I am not sure why the blackfriday slack package uses the former import path.
If I do govendor fetch github.com/russross/blackfriday/v2, the vendor.json is updated with the following entry - 
```
{
	"path": "github.com/russross/blackfriday/v2",
	"revision" : ""

},
```

borovskyav:
There is a bug in blackfriday-slack. Package ```github.com/russross/blackfriday/v2``` is unexists. Let's fork it, fix, test in Moira and PR in blackfriday-slack

aswinmprabhu:
@borovskyav I looked into this issue a bit more. I don't think this is a bug. Both blackfriday and blackfriday-slack are using go modules. This means that different versions of the package should have different import paths. The v2 branch of the blackfriday package does have a [go.mod](https://github.com/russross/blackfriday/blob/3e56bb68c8876389c631e9e318ce3c092a0906db/go.mod#L1) file that says the import path is github.com/russross/blackfriday/v2. This is module that is being used by blackfriday-slack. The README of blackfriday says that projects using older dependency tools like govendor should use the import path github.com/russross/blackfriday and specify that we are using v2. This can be done in govendor using the command `govendor fetch github.com/russross/blackfriday@=v2`. This does work when I tried it (instead of gopkg import path). So I think we will not be able to use blackfriday-slack without adopting go mod and semantic version importing. Maybe we can fork the package and use the path github.com/russross/blackfriday for our use? But I don't think they will accept a PR.

borovskyav:
Maybe we just can use previous commit without gomod? https://github.com/karriereat/blackfriday-slack/commit/1989d6c6eb352e231f903425588ea86c779b1978

aswinmprabhu:
> Maybe we just can use previous commit without gomod? karriereat/blackfriday-slack@1989d6c

This is a really good solution. I used it for the slack sender. But for the telegram sender there are differences like bold format is `**bold**` instead of `*bold*` as in slack. What should be done in this case? I guess the current custom solution works?

borovskyav:
It seemed to me that telegram supports the usual markdown that we have in the original description. Can you check it?

aswinmprabhu:
It doesn't support MD headers and italics
```
# header 1
## header 2
```

beevee:
Closing this PR, because this task was partly done in #403.

PR #372, opened 2019-05-25
Always use the latest patch version of Go

AlekSi:


beevee:
I'm not sure if it's a good idea to use different Go versions in Travis and Docker images. This PR changes only Travis config.

PR #384, opened 2019-06-16
Feature/prometheus 366

Pliner:


Tetrergeru:
Prometheus integration implemented here: #848

PR #386, opened 2019-06-20
Ignore sequential dots in metrics

Pliner:
```
$type.app.$cluster.acq-web.$host.s2724.$metric..cassandra.default.open-connections 5 1561029392
```


Pliner:
After discussion with @borovskyav we have decided to not to ignore dots, but to warn about a dropped metrics.

PR #387, opened 2019-06-24
Add discord sender

aswinmprabhu:


senders/discord/send.go
```diff
@@ -0,0 +1,151 @@
...
+	if len(plot) > 0 {
+		data.File = sender.buildPlot(plot)
+		data.Embed = &discordgo.MessageEmbed{
+			Image: &discordgo.MessageEmbedImage{
+				URL: "attachment://Plot.jpg",
+			},
+		}
+	}
+	sender.logger.Debugf("Calling discord with message %s", data.Content)
+	_, err := sender.session.ChannelMessageSendComplex(contact.Value, data)
```
kamaev: - How does channelID differ from channelName? And usage of which parameter could be more user-friendly?
- How will we be sending messages to users in private?

aswinmprabhu: [This](https://support.discordapp.com/hc/en-us/articles/206346498-Where-can-I-find-my-User-Server-Message-ID-) is the official article from discord support on how to get a channel ID. It also mentions how to get the channel ID of a private chat with an individual user. Using channel name is more user-friendly but it's not advisable as mentioned [here](https://stackoverflow.com/questions/41515134/discord-bot-cant-get-channel-by-name) and the discord API docs also don't mention anything about getting a channel by its name or getting a list of all channels and searching for it.

kamaev: So what's the arguments you see to use channel id instead of channel name?

kamaev: I am not saying that this is only right way even though i've already implemented such channel-name-to-id resolver (in my local repo)
I'm just interested in advantages and disadvantages of both methods

aswinmprabhu: 1. There can be multiple channels of the same name. 
2. There is no straight forward way (as far as I could find) of resolving a channel name to its unique ID. Can you please share the method you used?
The advantage of using channel names is obviously user-friendliness.

kamaev: > Can you please share the method you used?
Of course, No!

aswinmprabhu: > Of course, No!

I don't understand!? Why? By local repo did you mean closed source(I couldn't find anything in your fork)? Or do you want me to figure it out myself? :)
But the fact that multiple channels with the same name can exist should by default rule out the usage of channel names. Shouldn't it?
image

kamaev: > Or do you want me to figure it out myself?

Of course, Yes!

> multiple channels with the same name can exist

Take a look on our Telegram sender, it has the same entity called Group chats
You can see how do we store chat id's

Please take more time on this feature. 
Try to figure out how we can make life of our users easier in this particular case

senders/discord/init.go
```diff
@@ -0,0 +1,98 @@
...
+
+func (sender *Sender) runBot() {
+	err := sender.session.Open()
+	if err != nil {
+		sender.logger.Errorf("error creating a connection to discord: %s", err)
+	}
+	stop := make(chan os.Signal, 1)
+	signal.Notify(stop, syscall.SIGINT, syscall.SIGTERM, os.Interrupt)
+	<-stop
+	defer sender.session.Close()
```
borovskyav: Better use `sender.session.Close()` after `sender.session.Open()` (or after err checking, need to verify discordgo logic on Open session error). Main point: if your application will return from function before close method call, it newer close session.

borovskyav: As an example, you can see https://github.com/moira-alert/moira/blob/master/metric_source/remote/request.go#L31. In this example you must close Body before the function ends and `defer` exactly that keyword, that can guarantee this behavior. In your code `defer` is useless because you call in on end of fucntion.
Just read this article https://yourbasic.org/golang/defer/

senders/discord/init.go
```diff
@@ -0,0 +1,98 @@
...
+		default:
+			s.ChannelMessageSend(m.ChannelID, "Unsupported channel type")
+		}
+	}
+
+	return nil
+}
+
+func (sender *Sender) runBot() {
+	err := sender.session.Open()
```
borovskyav: We must care about the simple run bot. In most Moira installations users runs many Moira notifiers (replicas), the first reason is fault tolerance (if one of your replicas stopped working you anyway has efficient service). But, for example, telegram does not allow to run many replicas of telegram bot. You must check is discord allows this? If not then just use logic like in [telegram bot](https://github.com/moira-alert/moira/blob/master/senders/telegram/init.go#L79).

borovskyav: A simple way to check - just run two instances of notifier on local PC

senders/discord/init.go
```diff
@@ -0,0 +1,98 @@
...
+			sender.logger.Errorf("failed to handle incoming message: %s", err)
+		}
+	}
+	sender.session.AddHandler(handleMsg)
+
+	go sender.runBot()
+	return nil
+}
+
+func (sender *Sender) getResponse(s *discordgo.Session, m *discordgo.MessageCreate) error {
```
borovskyav: Are you really sure that discord needs the same logic with the bot as telegram? Telegram has stupid restrictions about user authorization (bot cannot send message to you if you didn't use !start command). I think you must check that discord has the same restrictions (or some other?)

aswinmprabhu: The !start command is being used just to associate the username/channel name with the unique channel ID. In the case of an individual user, any message would work but I used "!start" just like in telegram and discord channel to ensure consistency. Once the start command is used the user can use the username like "@aswinmprabhu" in the web UI instead of the channel ID. So I think it is necessary for discord as well but for a different reason.

senders/discord/init.go
aswinmprabhu: I tried running 2 instances of the notifier. What happens is that both instances of the bot get the events from discord and send messages to discord. Like,
image
This is not desirable. So I think adopting the logic of the telegram bot would be the way forward.

borovskyav: Yeah, I think telegam like solution is best in this situation!

senders/discord/init.go
borovskyav: But we can still use channel(user) ID as a unique ID for contact? Interesting, what does discord suggest doing in that cases. Hmmm.

borovskyav: There is some opportunity to see channel(user) ID from discord interface?

aswinmprabhu: Yes, channel ID is the only way to uniquely identify a channel (both text and private). As mentioned in [this comment](https://github.com/moira-alert/moira/pull/387#discussion_r298567230),
> [This](https://support.discordapp.com/hc/en-us/articles/206346498-Where-can-I-find-my-User-Server-Message-ID-) is the official article from discord support on how to get a channel ID. It also mentions how to get the channel ID of a private chat with an individual user. 
But @kamaev suggested switching to a telegram like solution that will enable users to just use the channel name or username in the web UI. This is more user-friendly but does involve issuing commands like "!start" to get it to work.

aswinmprabhu:
When I was trying to figure out why the telegram Init() tests are passing while the discord ones are failing in the latest build when they are the exact same tests, I noticed that the telegram sender never executes the `go sender.runTelebot()` statement because the `telebot.NewBot()` is making an API call that is failing and causing an early return with an error (while testing the Init() method). This is not the case with the discord sender, which executes the `runBot()` method and panics because the Database(referenced [here](https://github.com/aswinmprabhu/moira/blob/a22786d60de9b493cf940c5a6078acdbabd2045b/senders/discord/init.go#L106)) is nil.  Is this behavior while testing the telegram Init() method ideal? How should I deal with this issue in the discord sender?

borovskyav:
About `Init()` tests: yes, I know about this problem. The main reason why I don't fix this it is dependence form third party API. It adds one more point of failure of tests. In this reason, I just decided not to test scenarios with make telegram connections (simple example: telegram API is blocked in Russia and these tests will newer works).

borovskyav:
Yes, this with so sad that we can not test it. But I think we still can but this is will be so much spending of time to mock all these interfaces

senders/discord/init.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	return "", nil
+}
+
+func (sender *Sender) runBot() {
+	workerAction := func(stop <-chan struct{}) error {
+		err := sender.session.Open()
+		if err != nil {
+			sender.logger.Errorf("error creating a connection to discord: %s", err)
```
borovskyav: Let's return nil here, we need to exit from a function if we can not to open session

borovskyav:
You did a good job! I have a few comments that are more related to review https://github.com/moira-alert/moira/pull/370. Let's try to use Black Friday all the same because it covers a bunch of Markdown conversion scenarios, which we will cover very long for ourselves.

> I had tried using the blackfriday-slack package. It had a dependency issue. It seems to depend on github.com/russross/blackfriday/v2 while the README for the library says the import path is gopkg.in/russross/blackfriday.v2

Just use github.com/russross/blackfriday/v2. This is an acceptable solution for us since these two packages are the same in terms of golang.

Let's do this: we are now in a vimdzhim discord Sender in the master, and then fix the dependency on the package in https://github.com/moira-alert/moira/pull/370?

borovskyav:
I tried to rewrite the message building code a little:
```
func (sender *Sender) buildMessage(events moira.NotificationEvents, trigger moira.TriggerData, throttled bool) string {
  var buffer strings.Builder

  title := sender.buildTitle(events, trigger)
  titleLen := len([]rune(title))

  desc := sender.buildDescription(trigger)
  descLen := len([]rune(desc))

  eventsString := sender.buildEventsString(events, -1, throttled, trigger)
  eventsStringLen := len([]rune(eventsString))

  charsLeftAfterTitle := messageMaxCharacters - titleLen

  descNewLen, eventsNewLen := sender.calculateMessagePartsLength(charsLeftAfterTitle, descLen, eventsStringLen)

  if descLen > descNewLen {
    desc = desc[:descNewLen] + "...\n"
  }

  if eventsNewLen > eventsStringLen {
    eventsString = eventsString[:eventsNewLen]
  }

  buffer.WriteString(title)
  buffer.WriteString(desc)
  buffer.WriteString(eventsString)
  return buffer.String()
}

func (sender *Sender) calculateMessagePartsLength(maxChars, descLen, eventsLen int) (descNewLen int, eventsNewLen int) {
  if descLen+eventsLen <= messageMaxCharacters {
    return descLen, eventsLen
  }
  if descLen > maxChars/2 && eventsLen <= maxChars/2 {
    return maxChars - eventsLen, eventsLen
  }
  if eventsLen > maxChars/2 && descLen <= eventsLen/2 {
    return descLen, maxChars - descLen
  }
  return eventsLen/2 - 10, eventsLen / 2
}

func (sender *Sender) buildDescription(trigger moira.TriggerData) string {
  desc := trigger.Desc
  if trigger.Desc != "" {
    // Replace MD headers (## header text) with header text that telegram supports
    desc = mdHeaderRegex.ReplaceAllString(trigger.Desc, "$headertext")
    desc += "\n"
  }
  return desc
}

func (sender *Sender) buildTitle(events moira.NotificationEvents, trigger moira.TriggerData) string {
  state := events.GetSubjectState()
  tags := trigger.GetTags()
  return fmt.Sprintf("%s %s %s (%d)\n", state, trigger.Name, tags, len(events))
}
```

What advantages I see in it:
1. The `buildMessage` method has become cleaner, now there is a clear pipeline: build a header, message, event -> calculate the optimal length -> trim -> build the whole message
2. The logic of counting the circumcision of the message was put into a separate function, got rid of the copy-assy, the method can be reused in other senders

borovskyav:
Waiting for your answer.  Do we finish it in this PR or in # 370?

aswinmprabhu:
> Let's do this: we are now in merge discord Sender in the master, rebase MD-branch on master, and then fix the package dependency in #370?

Sure. But do we implement the changes you suggested for the message building algorithm in the PR of #370?

borovskyav:
Yes, you are right

borovskyav:
I will merge this PR as soon as possible

borovskyav:
For now, read the code above.  What do you think about it?  I think it turned out well, simply and clearly.

PR #389, opened 2019-07-04
Add pagerduty sender

aswinmprabhu:
This PR also includes the ImageStore interface that is used to represent image storages that are required to store the plot images for senders like pagerduty that cannot accept the image directly. It also includes an implementation of the interface for AWS S3 storage. It has to be configured with the notifer as shown below
```
# notifier.yml
....
notifier:
  ...
  image_store:
    type: "s3"
    access_key_id: "<access key id>"
    access_key: "<access key>"
    region: "ap-south-1"
    bucket: "moiratest"
  ...
```
The image store is used to store the image and link to the image is sent to pagerduty. The link has to be accessible freely with the link for pagerduty to receive the image. The unique key for the filename is generated in this PR using a uuid generator from google. The lack of configuration is not fatal to the pagerduty sender (only a warning is given). If no image store is configured, the message is sent without the image.


borovskyav:
I think it will be a good idea that notifier package will not contain image store configuration code. A simple example: yesterday one of our user write me "I want to see my pictures that stored in s3 in API by URL" (i don't know why, but this is just example).
Ok, now I need to reference notifier package from api. Bad case =(

Just lets make `InitImageStore` in cmd package. In this place, we can easily reuse it in other binaries like API.

borovskyav:
About code storing we try to use the following rules:
Every word in package name we separate with an underscore and use singular words. In this case, I will be rename imagestores to image_store.
We try to store config structures with code that this config configuring. In this case, I will create a new structure `Config` inside s3 package and use it in `init` method.
Let's remove `init` method from `ImageStore` interface. I think you see to sender interface when did it. For this case, I will be using a pattern like in database. In interface has only datastore methods, but `init` will be just like a constructor (like in javascript or c#, for example) who takes `Config` agrument

imagestores/s3/store.go
```diff
@@ -0,0 +1,34 @@
...
+	result, err := imageStore.uploader.Upload(uploadInput)
+	if err != nil {
+		return "", fmt.Errorf("error while uploading to s3: %s", err)
+	}
+
+	return result.Location, nil
+}
+
+func (imageStore *ImageStore) buildUploadInput(image []byte) *s3manager.UploadInput {
+	key := "moira-plots/" + uuid.New().String()
```
borovskyav: Can we use here "github.com/gofrs/uuid" package like in other Moira code? It will be a nice reason to remove the additional dependency.

borovskyav:
Let's change config format for `image_store`. A solution like not the best for this situation. For example, we may want to use multiple repositories for pictures. In this case, the current solution is not suitable. I propose such a solution:
```
# notifier.yml
....
notifier:
  ...
  image_store:
	s3:
    	access_key_id: "<access key id>"
    	access_key: "<access key>"
    	region: "ap-south-1"
    	bucket: "moiratest"
  ...
```
We already have problems with configs, where we did not adhere to this rule and in the next release we will have to change the format, and all users will correct their configs when updating, for example, in the settings of metrics sources. It usually hurts a lot.
General advice: always think about expanding your solution, how time-consuming and convenient it will be, both for you and for users.

aswinmprabhu:
@borovskyav I made all the changes you suggested. Now the user can have multiple image stores. The sender using image store will have to ID the store using a unique ID. Like,
```
...
notifier:
...
  image_store:
    s3:
      access_key_id: ""
      access_key: ""
      region: "ap-south-1"
      bucket: "moiratestaswin"
  senders:
    - type: pagerduty
      image_store: "s3"
...

**Comment by borovskyav** [comment]
By repositories, I mean different storage implementations, in this case, the ImageStore interface.
For example:
1. You have a `s3/ImageStore` implementation for storage images in s3
2. You have a WebDAV implementation (for example) for storage images on a WebDAV server

I advise you to read this book here https://www.amazon.com/Patterns-Enterprise-Application-Architecture-Martin/dp/0321127420. She describes such patterns as a repository well.

You can alse see such examples in Moira for example `MetricSource` interface.

**Comment by aswinmprabhu** [comment]
Yes, so I was right. Configuring different implementations(s3, webdav etc) of image stores and specifying which one to use for each sender is possible with the current implementation.

**Comment by borovskyav** [comment]
Yes, this is a good solution 👍

**Comment by borovskyav** [comment]
We can also return error on init sender if user choose here s3 storage and does not configure it

File: cmd/notifier/config.go [hunk 2]
```diff
@@ -36,8 +36,8 @@ type notifierConfig struct {
 	Timezone string `yaml:"timezone"`
 	// Format for email sender. Default is "15:04 02.01.2006". See https://golang.org/pkg/time/#Time.Format for more details about golang time formatting.
 	DateTimeFormat string `yaml:"date_time_format"`
-	// Image store config section.
-	ImageStore map[string]string `yaml:"image_store"`
+	// Image stores config section.
+	ImageStores map[string]map[string]string `yaml:"image_store"`
```
borovskyav: Let's make more stupid solution here. Just create some structures:
```
type imageStoreConfig struct {
	S3 s3Config `yaml:"s3"`
}

type s3Config struct {
	AccessKeyID string `yaml:"access_key_id"`
	AccessKey   string `yaml:"access_key"`
	Region      string `yaml:"region"`
	Bucket      string `yaml:"bucket"`
}
```

And replace `map[string]map[string]string` to `imageStoreConfig.
I do not see much point in using here such a flexible structure as `map[string]map[string]string` because you are hardcoded store type in `image_store.go` at the same time, it greatly complicates the code itself and its reading.

borovskyav: You can also add `enabled` field for an accurate transmission of whether the image store is enabled or not, or you can simply transfer the already written parameters: if configured, the image store is enabled, if not, then disabled.

imagestores/s3/init.go
```diff
@@ -10,6 +10,14 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 )
 
+// Config is the configuration structure for s3 image store
+type Config struct {
```
borovskyav: Let's move to separate file like another Moira configs.

notifier/config.go
```diff
@@ -10,7 +10,7 @@ type Config struct {
 	SendingTimeout   time.Duration
 	ResendingTimeout time.Duration
 	Senders          []map[string]string
-	ImageStore       map[string]string
+	ImageStores      map[string]map[string]string
```
borovskyav: Notifier config should not know about the settings of the image storage, he only needs to know what storages are. We use this pattern everywhere. For example, the same notifier does not know anything about RemoteStorage but at the same time, he uses it. Just look at the file `notifier/main.go` and how we use it.

senders/pagerduty/init.go
```diff
@@ -1,22 +1,28 @@
...
+	logger       moira.Logger
+	frontURI     string
+	location     *time.Location
 }
 
 // Init loads yaml config, configures the pagerduty client
 func (sender *Sender) Init(senderSettings map[string]string, logger moira.Logger, location *time.Location, dateTimeFormat string) error {
 	sender.frontURI = senderSettings["front_uri"]
+	sender.imageStoreID = senderSettings["image_store"]
+	if sender.imageStoreID == "" {
```
borovskyav: Let's just not save the picture if the type is not put. I think not every user wants to buy s3, but they will want to use pagerduty. Then with the release I will add this to the documentation.

senders/pagerduty/init.go
```diff
@@ -1,22 +1,28 @@
...
+		return fmt.Errorf("cannot read image_store from the config")
+	}
```
borovskyav: But it seems that if the user has put a image store, then we should definitely check that it is configured

notifier/config.go
borovskyav: For now, we can remove this field because it is no usages

senders/pagerduty/init.go
```diff
@@ -0,0 +1,30 @@
...
+	frontURI     string
+	location     *time.Location
+}
+
+// Init loads yaml config, configures the pagerduty client
+func (sender *Sender) Init(senderSettings map[string]string, logger moira.Logger, location *time.Location, dateTimeFormat string) error {
+	sender.frontURI = senderSettings["front_uri"]
+	sender.imageStoreID = senderSettings["image_store"]
+	if sender.imageStoreID == "" {
+		logger.Warningf("cannot read image_store from the config, will not be able to attach plot images to events")
```
borovskyav: All our log entries start with a capital letter. Look at the remaining places in the code where you added logging and fix, please.

senders/pagerduty/init.go
```diff
@@ -0,0 +1,30 @@
...
+	} else if !(sender.ImageStores[sender.imageStoreID].IsEnabled()) {
```
borovskyav: Sorry for bad comment about the right configuration of image store. I did not see that you return an error if store just not configured well. Can we remove this "IsEnabled" logic? Sorry =(

senders/pagerduty/send.go
```diff
@@ -0,0 +1,99 @@
...
+		Source:   "moira",
+		Details:  details,
+	}
+	event := pagerduty.V2Event{
+		RoutingKey: contact.Value,
+		Action:     "trigger",
+		Payload:    payload,
+	}
+
+	if len(plot) > 0 && sender.ImageStores[sender.imageStoreID].IsEnabled() {
```
borovskyav: I really do not like that you have a lot of places where you use the construction `sender.ImageStores[sender.imageStoreID]`. When writing new code, it can lead to bad consequences. Specifically, here you will fall into a nullPrinterError if the user has given a bad image store name.
Let's do a little differently. In the init method, we will check that such an image store is present in our map (if not, we will write a warning to the log) and if present, we will put in a separate field in Sender structure. Next, when trying to save a picture, we will check if there is any value in this field, if there is, then we will already save the picture.

senders/pagerduty/init.go
aswinmprabhu: I don't understand. Shouldn't we check if the image store is configured properly if an image store ID is specified for this sender? 
Earlier I had,
if "image_store" field doesn't exist, then return an error. (This will cause the sender to not work at all but we want it to work without the plotting capability in this case)

Now,
if "image_store" field doesn't exist, then warn the user and continue sender configuration.
if it exists, then check if it is configured, if not warn the user and continue.

The IsEnabled() is used to check if it is configured.

borovskyav: Oh, I understand what the problem is. You're right.

borovskyav: Let's act then in another way. It seems that putting this flag down is a leak of responsibility. Let's better make the Enabled flag private and will put it in the Init method after configuration.

senders/pagerduty/send.go
borovskyav: About this comment:

> Let's better make the Enabled flag private and will put it in the Init method after configuration.

Let's then set it up in a separate field if image store IsEnabled.

aswinmprabhu:
> We can fill in the links field with variables from the Trigger_URI if specified. href - Trigger_URI, text - TriggerName. And TriggerName can not be empty.

The go-pagerduty library doesn't support links. https://github.com/PagerDuty/go-pagerduty/blob/cf1437c7c8d62e7cf711e1a5c5bbb4c01f26e178/event_v2.go#L12

PR #398, opened 2019-07-23
Victorops Sender

aswinmprabhu:
Things to note - 
* A client sdk for the Victorops alert API has also been added in the `/api` directory of the sender due to lack of official and good third party sdks.
* The API manual for reference is https://help.victorops.com/knowledge-base/rest-endpoint-integration-guide/
* I could not find any char limits for the API. The state_message field which is used for the description and events string is quoted to have "high character limit" in the guide
* There is no way to attach images to the alerts for plots. One solution could be to just send the image links from the image store in text form (https://github.com/grafana/grafana/issues/6858#issuecomment-265559557)


senders/victorops/api/alert.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+// CreateAlert creates a new alert in the victorops timeline
+func (client *Client) CreateAlert(routingKey string, alert CreateAlertRequest) error {
+	if alert.MessageType == "" {
+		return fmt.Errorf("field MessageType cannot be empty")
+	}
+
+	body := new(bytes.Buffer)
+	err := json.NewEncoder(body).Encode(alert)
```
borovskyav: Let's use json.Marshal, as elsewhere in Moira, [example](https://github.com/moira-alert/moira/blob/master/metric_source/remote/response.go#L28) (this is the reverse example, which is json.Unmarshal response body)

senders/victorops/send.go
```diff
@@ -0,0 +1,112 @@
...
+	switch events[len(events)-1].State {
+	case moira.StateERROR:
+		messageType = api.Critical
+	case moira.StateEXCEPTION:
+		messageType = api.Critical
+	case moira.StateWARN:
+		messageType = api.Warning
+	case moira.StateOK:
+		messageType = api.Recovery
+	case moira.StateNODATA:
```
borovskyav: For us, the NODATA state is as bad as the error state, let's use here api.Critical

senders/victorops/api/alert.go
```diff
@@ -0,0 +1,63 @@
...
+	if err != nil {
+		return fmt.Errorf("error while encoding json: %s", err)
+	}
+	req, err := http.NewRequest("POST", fmt.Sprintf("%s/%s", client.routingURL, routingKey), body)
```
borovskyav: As I understand it in url there are three parts.
First: the constant https://alert.victorops.com/integrations/generic/20131114/alert.
Second: API key that you need to pass in the sender settings
Third: routingKey that the user passes

Let's get this to look like this?

senders/victorops/send.go
```diff
@@ -0,0 +1,112 @@
...
+	title := string(events.GetSubjectState())
+
+	tags := trigger.GetTags()
+	if tags != "" {
+		title += " " + tags
+	}
+
+	triggerURI := trigger.GetTriggerURI(sender.frontURI)
+	if triggerURI != "" {
+		title += fmt.Sprintf(" %s|%s", triggerURI, trigger.Name)
```
borovskyav: Are you sure that link formatting works here? You can throw off the documentation, I did not find it =(

borovskyav: If it does not work, then we can insert a link to a trigger in a free field (for example, to a trigger_url)

borovskyav: we can do the same with reference to the picture, create a custom field, for example image_url

senders/victorops/api/alert.go
```diff
@@ -0,0 +1,63 @@
...
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+)
+
+// CreateAlertRequest the API request to be made to
+// create a victorops alert
+type CreateAlertRequest struct {
```
borovskyav: There are a lot of additional fields in [grafana integration](https://github.com/grafana/grafana/blob/d8736a25475d146bd7111bddc5739e1380a1b0d3/pkg/services/alerting/notifiers/victorops.go#L104) and in a [third-party client](https://github.com/chrissnell/victorops-go/blob/master/client.go#L17), let's start transferring them too.

borovskyav: I would pass these fields:
- timestamp: time.Now().Unix()
- entity_id: TriggerID (It seemed to me that this field is required. Have you checked if the integration works?)
- monitoring_tool: Moira

senders/victorops/api/alert.go
aswinmprabhu: I am not sure whether the URL is a constant or unique to the user. That is why I designed it in such a way that the user will pass the entire URL and the api key via the sender settings, ie, https://alert.victorops.com/integrations/generic/20131114/alert/<api_key>. This is called the routingURL in the code. Thee routing key is then obtained from the web UI.

senders/victorops/send.go
```diff
@@ -0,0 +1,130 @@
...
+	err := sender.client.CreateAlert(contact.Value, createAlertRequest)
+	if err != nil {
+		return fmt.Errorf("error while sending alert to victorops: %s", err)
+	}
+	return nil
+}
+
+func (sender *Sender) buildCreateAlertRequest(events moira.NotificationEvents, trigger moira.TriggerData, throttled bool, plot []byte, time int64) api.CreateAlertRequest {
+
+	var messageType api.MessageType
```
borovskyav: It seems we have problems with determining the type of message here and in PagerDuty.
See how it is done in [pushover](https://github.com/moira-alert/moira/blob/master/senders/pushover/pushover.go#L57)
But I think that this decision is unnecessary. It seems the easiest way to use a ```GetSubjectStat``` here and use this state in calculating the correct message type. Let's redo it here and in PagerDuty

senders/victorops/api/alert.go
borovskyav: Well, even if one of us is wrong, I do not see big problems in this decision =)

senders/victorops/send_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"testing"
+	"time"
+
+	"github.com/moira-alert/moira/senders/victorops/api"
+
+	"github.com/moira-alert/moira"
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+type MockImageStore struct {
```
borovskyav: For some reason I did not immediately notice how you made this decision in pagerduty =(
For these stubs we use gomock. You can add your interface in the files generate_mocks.(cmd and sh) and execute then. Let's try to generate mock for image store and replace them.

borovskyav:
Good job! Please dont forget to add new sender to web =)

PR #405, opened 2019-08-05
Opsgenie Sender

aswinmprabhu:


senders/opsgenie/send.go
```diff
@@ -0,0 +1,152 @@
...
+	htmlDescLen := len([]rune(htmlDesc))
+	charsForHTMLTags := htmlDescLen - len([]rune(desc))
+
+	eventsString := sender.buildEventsString(events, -1, throttled)
+	eventsStringLen := len([]rune(eventsString))
+
+	descNewLen, eventsNewLen := senders.CalculateMessagePartsLength(msgLimit, htmlDescLen, eventsStringLen)
+
+	if htmlDescLen != descNewLen {
+		desc = desc[:descNewLen-charsForHTMLTags] + "...\n"
```
borovskyav: I don't understand this logic =( 
Can you explain to me, please? =)
I really don't understand why you additionally cut desc to charsForHTMLTags

senders/opsgenie/send.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	return createAlertRequest
+}
+
+func (sender *Sender) buildMessage(events moira.NotificationEvents, throttled bool, trigger moira.TriggerData) string {
+	var message strings.Builder
+
+	desc := trigger.Desc
+	htmlDesc := string(blackfriday.Run([]byte(desc)))
```
borovskyav: Сan opsgenie render html tags? I didn't find it in api documentation.

borovskyav: If opsgenie can not parse html tags, maybe we can reformat it to simple text? Just check is blackfriday has this feature?

aswinmprabhu: Yes it can. I did verify the integration. Check out https://docs.opsgenie.com/docs/alerts-and-alert-fields.

senders/opsgenie/send.go
```diff
@@ -0,0 +1,152 @@
...
+)
+
+// SendEvents sends the events as an alert to opsgenie
+func (sender *Sender) SendEvents(events moira.NotificationEvents, contact moira.ContactData, trigger moira.TriggerData, plot []byte, throttled bool) error {
+	createAlertRequest := sender.makeCreateAlertRequest(events, contact, trigger, plot, throttled)
+	_, err := sender.client.Create(context.Background(), createAlertRequest)
+	if err != nil {
+		return fmt.Errorf("error while creating alert: %s", err)
+	}
+	if err != nil {
```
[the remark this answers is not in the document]
aswinmprabhu: Yes I did test it out. The 1st error check was left over from when I was playing around with the code from their docs :)
Removed it!

senders/opsgenie/send.go
aswinmprabhu: Because we need to leave space for the html tags. After trimming the desc based on descNewLen we are again converting MD to html using blackfriday. This will add more chars to the desc. charsForHTMLTags is a rough estimate of how many chars will be required for the tags. So we trim the desc to descNewLen - charsForHTMLTags so that when the html tags are added the final len will be less than descNewLen. Of course some chars are also lost in the html conversion because the MD syntax is stripped and replaced with HTML tags. But this conversion cannot be precisely calculated. charsForHTMLTags gives us a safe figure for the chars that might be required.

senders/opsgenie/init.go
```diff
@@ -0,0 +1,58 @@
...
+
+// Init initializes the opsgenie sender
+func (sender *Sender) Init(senderSettings map[string]string, logger moira.Logger, location *time.Location, dateTimeFormat string) error {
+	var ok bool
+
+	if sender.apiKey, ok = senderSettings["api_key"]; !ok {
+		return fmt.Errorf("cannot read the api_key from the sender settings")
+	}
+
+	sender.imageStoreID, ok = senderSettings["image_store"]
```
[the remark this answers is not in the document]
aswinmprabhu: Yes. All senders using image store will have to have this snippet. I moved it to the senders package so that both opsgenie and pagerduty can use it. Also added tests!

PR #412, opened 2019-08-14
Fix pagerduty timestamp error

aswinmprabhu:


senders/pagerduty/send_test.go
```diff
@@ -48,7 +48,7 @@ func TestBuildEvent(t *testing.T) {
 				Summary:   "NODATA [tag1][tag2]",
 				Severity:  "warning",
 				Source:    "moira",
-				Timestamp: "150000000",
+				Timestamp: time.Unix(150000000, 0).Format(time.RFC3339),
```
beevee: I think, a more proper test would be to match against `1974-10-03T02:40:00Z`. In your current code you are checking that `time.Unix(...).Format(...)` produces the same result when called from different places. @borovskyav what do you think?

borovskyav: I think @beevee is right. Thus, we verify that that formatting method satisfies pagerduty validations. This may work, for example, if in some version of go these methods start working differently (which of course will be very strange)

aswinmprabhu:
Checkout [this](https://play.golang.org/p/90QyqkrjCzo) code.
Running it in the playground gives me output
```
Hello, playground
1974-10-03T02:40:00Z
```
But on my local machine it gives me the output 
```
Hello, playground
1974-10-03T08:10:00+05:30
```
image

PR #432, opened 2019-10-01
Msteams support

imavroukakis:
Adds support for Microsoft Teams, via inbound channel WebHooks

* Add a delivery channel using the MSTeams Incoming Webhook 
image
* Assign the delivery channel to a Subscription (graphs are not yet supported)
image

Wait for the alerts to roll in the channel

image

### DIFF
--- a/senders/msteams/datatypes.go
+++ b/senders/msteams/datatypes.go
@@ -1,5 +1,9 @@
 package msteams
 
+import (
+	"fmt"
+)
+
 /*
 Fact models a fact in a MessageCard, contains a timestamp and trigger data
  {
@@ -105,3 +109,15 @@ type MessageCard struct {
 	Sections        []Section `json:"sections"`
 	PotentialAction []Actions `json:"potentialAction,omitempty"`
 }
+
+type ErrTeamsError struct {
+	error       string
+	description string
+	err         error
+}
+
+func (e *ErrTeamsError) Error() string {
+	return fmt.Sprintf("%s : %s", e.description, e.error)
+}
+
+func (e *ErrTeamsError) Unwrap() error { return e.err }
--- a/senders/msteams/datatypes.go
+++ b/senders/msteams/datatypes.go
@@ -1,7 +1,7 @@
 package msteams
 
 /*
-Represents a Fact in a MessageCard, contains a timestamp and trigger data
+Fact models a fact in a MessageCard, contains a timestamp and trigger data
  {
 		"name": "10:45",
     "value": "someServer = 0.11 (NODATA to WARN)"
@@ -13,7 +13,7 @@ type Fact struct {
 }
 
 /*
-Represents a Section in a MessageCard, contains Facts and the Trigger description
+Section models a section in a MessageCard, contains Facts and the Trigger description
  {
 		"activityTitle": "Description",
 		"activityText": "A trigger description",
@@ -32,7 +32,7 @@ type Section struct {
 }
 
 /*
-Represents an OpenURITarget data in a MessageCard,creates a clickable target back to the trigger URI
+OpenURITarget creates a clickable target back to the trigger URI in a MessageCard
  {
 		"os": "default",
     "value": "http://moira.tld/trigger/ABCDEF-GH"
@@ -44,7 +44,7 @@ type OpenURITarget struct {
 }
 
 /*
-Represents possible actions in a MessageCard, limited to OpenURI actions
+Actions models possible actions in a MessageCard, currently limited to OpenURI actions
  {
 		"@type": "OpenUri",
     "name": "Open in Moira"
@@ -63,7 +63,7 @@ type Actions struct {
 }
 
 /*
-Represents an MSTeams compatible MessageCard
+MessageCard models an MSTeams compatible MessageCard
  {
 		"@context": "https://schema.org/extensions",
     "@type": "MessageCard",
--- a/senders/msteams/datatypes.go
+++ b/senders/msteams/datatypes.go
@@ -1,28 +1,101 @@
 package msteams
 
+/*
+Represents a Fact in a MessageCard, contains a timestamp and trigger data
+ {
+		"name": "10:45",
+    "value": "someServer = 0.11 (NODATA to WARN)"
+ }
+*/
 type Fact struct {
 	Name  string `json:"name"`
 	Value string `json:"value"`
 }
 
+/*
+Represents a Section in a MessageCard, contains Facts and the Trigger description
+ {
+		"activityTitle": "Description",
+		"activityText": "A trigger description",
+    "facts": [
+			 {
+					"name": "10:45",
+					"value": "someServer = 0.11 (NODATA to WARN)"
+			 }
+		]
+ }
+*/
 type Section struct {
 	ActivityTitle string `json:"activityTitle"`
 	ActivityText  string `json:"activityText"`
 	Facts         []Fact `json:"facts"`
 }
 
-type OpenUriTarget struct {
+/*
+Represents an OpenURITarget data in a MessageCard,creates a clickable target back to the trigger URI
+ {
+		"os": "default",
+    "value": "http://moira.tld/trigger/ABCDEF-GH"
+ }
+*/
+type OpenURITarget struct {
 	Os  string `json:"os"`
-	Uri string `json:"uri"`
+	URI string `json:"uri"`
 }
 
-//limited to OpenURI actions
+/*
+Represents possible actions in a MessageCard, limited to OpenURI actions
+ {
+		"@type": "OpenUri",
+    "name": "Open in Moira"
+		"targets": [
+			{
+				"os": "default",
+				"value": "http://moira.tld/trigger/ABCDEF-GH"
+ 			}
+		]
+ }
+*/
 type Actions struct {
 	Type    string          `json:"@type"`
 	Name    string          `json:"name"`
-	Targets []OpenUriTarget `json:"targets"`
+	Targets []OpenURITarget `json:"targets"`
 }
 
+/*
+Represents an MSTeams compatible MessageCard
+ {
+		"@context": "https://schema.org/extensions",
+    "@type": "MessageCard",
+		"summary": "Moira Alert"
+		"title" : "WARN Trigger Name [tag1]"
+		"themeColor": "ffa500"
+		"sections": [
+			 {
+					"activityTitle": "Description",
+					"activityText": "A trigger description",
+					"facts": [
+						 {
+								"name": "10:45",
+								"value": "someServer = 0.11 (NODATA to WARN)"
+						 }
+					]
+			 }
+		]
+		"potentialAction": [
+			{
+				"@type": "OpenUri",
+				"name": "Open in Moira"
+				"targets": [
+					{
+						"os": "default",
+						"value": "http://moira.tld/trigger/ABCDEF-GH"
+					}
+				]
+			}
+		]
+ }
+*/
 type MessageCard struct {
 	Context         string    `json:"@context"`
 	MessageType     string    `json:"@type"`
--- a/senders/msteams/datatypes.go
+++ b/senders/msteams/datatypes.go
@@ -50,7 +50,7 @@ type OpenURITarget struct {
 }
 
 /*
-Actions models possible actions in a MessageCard, currently limited to OpenURI actions
+Action models possible action in a MessageCard, currently limited to OpenURI actions
  {
 		"@type": "OpenUri",
     "name": "Open in Moira"
--- a/senders/msteams/datatypes.go
+++ b/senders/msteams/datatypes.go
@@ -109,5 +109,5 @@ type MessageCard struct {
 	ThemeColor      string    `json:"themeColor"`
 	Title           string    `json:"title"`
 	Sections        []Section `json:"sections"`
-	PotentialAction []Actions `json:"potentialAction,omitempty"`
+	PotentialAction []Action `json:"potentialAction,omitempty"`
 }
--- a/senders/msteams/datatypes.go
+++ b/senders/msteams/datatypes.go
@@ -1,11 +1,8 @@
 package msteams
 
-import "github.com/moira-alert/moira"
-
 type Fact struct {
 	Name  string `json:"name"`
 	Value string `json:"value"`
-	state moira.State
 }
 
 type Section struct {
@@ -33,5 +30,5 @@ type MessageCard struct {
 	ThemeColor      string    `json:"themeColor"`
 	Title           string    `json:"title"`
 	Sections        []Section `json:"sections"`
-	PotentialAction []Actions `json:"potentialAction"`
+	PotentialAction []Actions `json:"potentialAction,omitempty"`
 }
--- a/senders/msteams/datatypes.go
+++ b/senders/msteams/datatypes.go
@@ -1,5 +1,11 @@
 package msteams
 
+const Green = "008000"
+const Orange = "ffa500"
+const Red = "ff0000"
+const Black = "000000"
+const White = "ffffff"
+
 /*
 Fact models a fact in a MessageCard, contains a timestamp and trigger data
  {
--- a/senders/msteams/datatypes.go
+++ b/senders/msteams/datatypes.go
@@ -62,7 +62,7 @@ Actions models possible actions in a MessageCard, currently limited to OpenURI a
 		]
  }
 */
-type Actions struct {
+type Action struct {
 	Type    string          `json:"@type"`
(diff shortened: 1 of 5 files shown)


senders/msteams/msteams.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+// Init initialises settings required for full functionality
+func (sender *Sender) Init(senderSettings map[string]string, logger moira.Logger, location *time.Location, dateTimeFormat string) error {
+	sender.logger = logger
+	sender.location = location
+	sender.frontURI = senderSettings["front_uri"]
+	sender.client = &http.Client{
+		Timeout:   time.Duration(30) * time.Second,
+		Transport: &http.Transport{DisableKeepAlives: true},
```
litleleprikon: Why keep alive is disabled here? As I can understand ms teams have only one domain and keeping connect will not raise memory usage depending on amount of alerts.

imavroukakis: good point!

imavroukakis: how about we make it configurable?

imavroukakis: I'm thinking something like this
```
disableKeepAlives, err := strconv.ParseBool(senderSettings["disable_keepalives"])
if err != nil {
	disableKeepAlives = true
}
sender.client = &http.Client{
	Timeout:   time.Duration(30) * time.Second,
	Transport: &http.Transport{DisableKeepAlives: disableKeepAlives},
}
```

litleleprikon:
@imavroukakis Can you please rebase on the fresh master, fix conflicts and push this branch again? As we moved to go modules your PR is using outdated dependencies. Especially be aware of changed path for blackfriday library.

Thank you!

imavroukakis:
> @imavroukakis Can you please rebase on the fresh master, fix conflicts and push this branch again? As we moved to go modules your PR is using outdated dependencies. Especially be aware of changed path for blackfriday library.
> 
> Thank you!

You got it

senders/msteams/msteams.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+func getColourForState(state moira.State) string {
+	switch state := state; state {
+	case moira.StateOK:
+		return "008000" //green
+	case moira.StateWARN:
+		return "ffa500" //orange
+	case moira.StateERROR:
+		return "ffa500" //red
```
litleleprikon: Should be a red color code

imavroukakis: fixed, thanks!

imavroukakis:
@litleleprikon good for another pass?

senders/msteams/msteams.go
```diff
@@ -0,0 +1,213 @@
...
+	sender.logger = logger
+	sender.location = location
+	sender.frontURI = senderSettings["front_uri"]
+	disableKeepAlives, err := strconv.ParseBool(senderSettings["disable_keepalives"])
+	if err != nil {
+		disableKeepAlives = false
+	}
+	sender.client = &http.Client{
+		Timeout:   time.Duration(30) * time.Second,
+		Transport: &http.Transport{DisableKeepAlives: disableKeepAlives},
```
litleleprikon: I still don't see the cases where we need this functionality/ But maybe I am wrong. @beevee what do you think about it?

beevee: @imavroukakis can you please explain, why would anybody want to disable keepalives? Is there a reason (maybe a bug in MS Teams API or something)?

imavroukakis: No bug that I'm aware of @beevee - I don't have a strong argument for or against this, except this: You _could_ argue, that with keep-alive turned on, you could eventually reach a connection limit on the MS server. Since this is an unknown value, it should at least be made configurable for users with a lot of webhooks. I think, enabling keep-alives by default is a sane strategy, with the option to turn them off if connection capacity issues are encountered.

beevee: As far as I understand, enabling keepalives will result in one (?) established connection per one notifier instance. For all practical purposes, there is no need to keep more than 2 or 3 notifier instances running. We only have one, since our Kubernetes cluster handles automatic failover, and high load is not an issue here.

If you ask me, I'd say that adding clutter to configuration files and code is not worth having even 3 long-running connections to MS Teams server.

imavroukakis: I'm happy to take it out then :)

imavroukakis: Done with https://github.com/moira-alert/moira/pull/432/commits/6f4045397f0f583d97fdadf299518e8ac8177ea4

senders/msteams/msteams.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+// Init initialises settings required for full functionality
+func (sender *Sender) Init(senderSettings map[string]string, logger moira.Logger, location *time.Location, dateTimeFormat string) error {
+	sender.logger = logger
+	sender.location = location
+	sender.frontURI = senderSettings["front_uri"]
+	sender.client = &http.Client{
+		Timeout:   time.Duration(30) * time.Second,
+		Transport: &http.Transport{DisableKeepAlives: false},
```
litleleprikon: Hi @imavroukakis! Thank you for the changes you made here! Let's remove this redundant transport declaration to not distort anyone who will read this code later.

This declaration is redundant because by default `Transport` will be `nil` here and `http.Client` will use default transport as written here https://golang.org/src/net/http/client.go#L193.

The default transport is decalred here https://golang.org/src/net/http/transport.go?#L42 and `DisableKeepAlives` is not set to any value which leads us to `DisableKeepAlives` be set to false.

imavroukakis: @litleleprikon 👍

imavroukakis: @litleleprikon done

litleleprikon: @imavroukakis Perfect! Thank You!

litleleprikon:
@beevee can you please check again that everything is alright and merge?

senders/msteams/msteams.go
```diff
@@ -0,0 +1,207 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"github.com/moira-alert/moira"
+	"github.com/russross/blackfriday/v2"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+)
```
beevee: Imports are not sorted properly here. I don't know why our linter misses this :-(

Please, sort imports using `goimports` (https://godoc.org/golang.org/x/tools/cmd/goimports).

senders/msteams/msteams_test.go
```diff
@@ -0,0 +1,374 @@
+package msteams
+
+import (
+	"github.com/moira-alert/moira/logging/go-logging"
+	"testing"
+	"time"
+
+	"github.com/moira-alert/moira"
+	. "github.com/smartystreets/goconvey/convey"
+)
```
beevee: Imports also not sorted here.

senders/msteams/msteams.go
```diff
@@ -0,0 +1,207 @@
...
+	err := sender.isValidWebhookURL(contact.Value)
+	if err != nil {
+		return fmt.Errorf("invalid msteams webhook url: %s", err.Error())
+	}
+
+	request, err := sender.buildRequest(events, contact, trigger, plot, throttled)
+
+	if request != nil {
+		defer request.Body.Close()
+	}
```
beevee: As per documentation, `The request Body, if non-nil, will be closed by the underlying Transport, even on errors.` So we don't need this.

I understand that you copied this code from our own webhooks, but we can make it better (and also raise an issue to make webhooks code better).

senders/msteams/msteams.go
```diff
@@ -0,0 +1,207 @@
...
+
+	response, err := sender.client.Do(request)
+	if response != nil {
+		defer response.Body.Close()
+	}
+
+	if err != nil {
+		return fmt.Errorf("failed to perform request: %s", err.Error())
+	}
```
beevee: I'd also rewrite this to be more in line with Go http module documentation:
```
if err != nil {
  return fmt.Errorf("failed to perform request: %s", err.Error())
}
defer response.Body.Close()
```

I see one more problem here. As per Go http module docs, `A non-2xx status code doesn't cause an error`. So this code will emit an error log on timeout or context cancellation, but no error on 500 response code from MS Teams API. I think we should check for `response.StatusCode`.

And another problem. As per docs, `If the Body is not both read to EOF and closed, the Client's underlying RoundTripper (typically Transport) may not be able to re-use a persistent TCP connection to the server for a subsequent "keep-alive" request`. So we have to read the body to EOF, not just close.

See https://golang.org/pkg/net/http/#Client.Do

imavroukakis:
@beevee done with your review changes under https://github.com/moira-alert/moira/pull/432/commits/3e6f9ce135aa0a89da085d4858b1098b25b7ae43

senders/msteams/msteams.go
```diff
@@ -0,0 +1,219 @@
...
+					Os:  "default",
+					URI: uri,
+				},
+			},
+		})
+	}
+	return MessageCard{
+		Context:     "http://schema.org/extensions",
+		MessageType: "MessageCard",
+		Summary:     "Moira Alert",
```
litleleprikon: Probably will be better to extract values to constants

senders/msteams/msteams.go
```diff
@@ -0,0 +1,219 @@
...
+
+	if err != nil {
+		return fmt.Errorf("failed to perform request: %s", err.Error())
+	}
+	defer response.Body.Close()
+
+	// read the entire response as required by https://golang.org/pkg/net/http/#Client.Do
+	body, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return fmt.Errorf("failed to decode response: %s", err.Error())
```
litleleprikon: I would like to see error wrapping using go 1.13 error values here and elsewhere  https://github.com/golang/go/wiki/ErrorValueFAQ

senders/msteams/msteams.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	sender.logger.Debugf("%s\n", requestBody)
+	request, err := http.NewRequest("POST", requestURL, bytes.NewBuffer(requestBody))
+	if err != nil {
+		return request, err
+	}
+	headers := map[string]string{
+		"User-Agent":   "Moira",
+		"Content-Type": "application/json",
```
litleleprikon: Also can be extracted to constants

senders/msteams/msteams.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+func (sender *Sender) isValidWebhookURL(webhookURL string) error {
+	// basic URL check
+	_, err := url.Parse(webhookURL)
+	if err != nil {
+		return err
+	}
+	// only pass MS teams webhook URLs
+	hasPrefix := strings.HasPrefix(webhookURL, "https://outlook.office.com/webhook/")
```
litleleprikon: Also will be better to extract to const

senders/msteams/msteams.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+func (sender *Sender) buildMessage(events moira.NotificationEvents, trigger moira.TriggerData, throttled bool) MessageCard {
+
+	title, uri := sender.buildTitleAndURI(events, trigger)
+	var triggerDescription string
+	if trigger.Desc != "" {
+		triggerDescription = string(blackfriday.Run([]byte(trigger.Desc)))
+	}
+	facts := sender.buildEventsFacts(events, -1, throttled)
```
litleleprikon: What is the purpose to set `maxEvents` here as hardcoded magic value? I believe that this value should be set in configuration

senders/msteams/msteams.go
```diff
@@ -0,0 +1,219 @@
...
+		return fmt.Errorf("failed to decode response: %s", err.Error())
+	}
+
+	//handle non 2xx responses
+	if response.StatusCode >= http.StatusBadRequest && response.StatusCode <= http.StatusNetworkAuthenticationRequired {
+		return fmt.Errorf("server responded with a non 2xx code: %d", response.StatusCode)
+	}
+
+	responseData := string(body)
+	if responseData != "1" {
```
litleleprikon: `"1"` should be a const

imavroukakis:
@litleleprikon changes ready for your review

senders/msteams/datatypes.go
```diff
@@ -105,3 +109,15 @@ type MessageCard struct {
...
+	error       string
+	description string
+	err         error
+}
+
+func (e *ErrTeamsError) Error() string {
+	return fmt.Sprintf("%s : %s", e.description, e.error)
+}
+
+func (e *ErrTeamsError) Unwrap() error { return e.err }
```
litleleprikon: Hey @imavroukakis there is no need to implement this error structures by yourself. My comment was only about changing format string to use `%w` placeholders instead of `%s`. The default errors returned by `errors.New` and `fmt.Errorf` are compatible with Go 1.13 error values. You can check example here https://play.golang.org/p/_2gKhr5mt05

Let's just remove this struct and use built-in errors generated by `errors.New` and `fmt.Errorf`

imavroukakis: So, for returning an error with a status code, this is acceptable?
```
if response.StatusCode >= http.StatusBadRequest && response.StatusCode <= http.StatusNetworkAuthenticationRequired {
		return fmt.Errorf("server responded with a non 2xx code: %d", response.StatusCode)

}
```

litleleprikon: Yeah, this is enough. If you will handle this error somewhere on higher levels, you will be able just wrap it with `fmt.Errorf("some additional info: %w")`

Another one note: Please check that all errors do not contains capitalised strings https://github.com/golang/go/wiki/CodeReviewComments#error-strings

senders/msteams/msteams.go
```diff
@@ -0,0 +1,234 @@
...
+
+	messageCard := sender.buildMessage(events, trigger, throttled)
+	requestURL := contact.Value
+	requestBody, err := json.Marshal(messageCard)
+	if err != nil {
+		return nil, err
+	}
+
+	sender.logger.Debugf("%s\n", requestBody)
+	request, err := http.NewRequest("POST", requestURL, bytes.NewBuffer(requestBody))
```
litleleprikon: `http.MethodPost` will be better here instead of `"POST"` string

senders/msteams/msteams.go
imavroukakis: aaah I fell through the D'OH tree and hit every branch on the way down, thanks for spotting this.

senders/msteams/msteams.go
```diff
@@ -0,0 +1,240 @@
...
+func (sender *Sender) buildRequest(events moira.NotificationEvents, contact moira.ContactData, trigger moira.TriggerData, plot []byte, throttled bool) (*http.Request, error) {
+
+	messageCard := sender.buildMessage(events, trigger, throttled)
+	requestURL := contact.Value
+	requestBody, err := json.Marshal(messageCard)
+	if err != nil {
+		return nil, err
+	}
+
+	sender.logger.Debugf("%s\n", requestBody)
```
litleleprikon: Should this debug messages appear here? If they should then we need to make them more verbose and add at least a bit of context instead of just plain JSON.

imavroukakis: I think we can forego this debug message in favour of the one at L161

`sender.logger.Debugf("%s %s '%s'", request.Method, request.URL.String(), string(requestBody))`

imavroukakis: and maybe augment it a bit to indicate that it's an MS Teams message

litleleprikon: Yeah, I think it is good point

senders/msteams/msteams_test.go
```diff
@@ -0,0 +1,447 @@
...
+								Value: "```Metric = 123 (OK to NODATA)```",
+							},
+							{
+								Name:  "Warning",
+								Value: "Please, *fix your system or tune this trigger* to generate less events.",
+							},
+						},
+					},
+				},
+				PotentialAction: []Actions{
```
litleleprikon:
```suggestion
				PotentialAction: []Action{
```

imavroukakis:
@litleleprikon done, apologies about the initial singular commits, the UI tripped me up a bit until I figured out where the batch commits are done.

imavroukakis:
> > > I will wait another one approve from @beevee. He will merge I believe
> > 
> > 
> > Thanks Emil!
> 
> You are welcome! We appreciate your involve in Moira alerting system and hope to see your improvements again.

Definitely, it was good fun and it taught me a lot.

imavroukakis:
@beevee good to go?

PR #433, opened 2019-10-01
discrete data volume for redis

imavroukakis:
use a discrete,named volume for redis data

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,6 +3,8 @@ version: "3.7"
 services:
   redis:
     image: redis:alpine
+    volumes:
+     - data:/data
 
   graphite:
     image: graphiteapp/graphite-statsd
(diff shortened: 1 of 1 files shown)


beevee:
Is there any reason to name this volume differently from https://github.com/moira-alert/docker-compose/blob/9f01fc34713e0d84b15a163c3e57476d7b14f6a7/docker-compose.yml#L6-L7 ?

imavroukakis:
Good question @beevee ,https://github.com/moira-alert/docker-compose/blob/9f01fc34713e0d84b15a163c3e57476d7b14f6a7/docker-compose.yml#L6-L7 should probably be renamed to how this is done here.

This PR generates a volume named `moira_data`, the other one `moira_moira-data`. I find `moira_data` to be better, unless, we further quantify it and name it `redis-data` to get `moira_redis-data`. What do you think?

beevee:
Okay, I'm going to merge this and change the other repo accordingly.

PR #434, opened 2019-10-01
build: Move to go modules

### LINKED ISSUE #423: Migrate to Go modules
I tried to migrate Moira from govendor to go modules in Go 1.12, but conversion failed due to some unsupported features of govendor (probably https://github.com/golang/go/issues/25556, not sure). We need to check again if it's possible for Go 1.13.

litleleprikon:
Drop usage of govendor as dependency management tool in favor to
go modules.

Closes #423


litleleprikon:
The errors in build seems to be due to old version of go used in our CI pipeline. The pipeline is using go 1.12.6 however my setup is go 1.13.1. I think the solution is to move CI to go 1.13.1.

beevee:
Also here: https://moira.readthedocs.io/en/latest/development/notifier.html

litleleprikon:
> > move CI to go 1.13.1
> 
> Yeah, just do it :-)
> 
> Looks like you forgot all the Dockerfiles, and also README. Grepping the repository should return zero instances of word `govendor` after this is merged.

Fixed

Dockerfile.api
```diff
@@ -1,10 +1,8 @@
 FROM golang:1.13.1 as builder
 
-RUN go get github.com/kardianos/govendor
-
 COPY ./vendor/vendor.json /go/src/github.com/moira-alert/moira/vendor/vendor.json
```
beevee: This line need attention. Also, look for `vendor.json` usages elsewhere, and fix accordingly.

Makefile
```diff
@@ -23,17 +24,16 @@ default: test build
 
 .PHONY: prepare
 prepare:
-	go get -u github.com/kardianos/govendor
-	govendor sync
+	go get ./...
```
beevee: Is this really necessary? Can we get rid of `prepare` step entirely with Go modules?

README.md
```diff
@@ -28,8 +28,7 @@ Other installation methods are available, see [documentation](https://moira.read
 To build and run tests, first get all dependencies:
 
 ```bash
-go get github.com/kardianos/govendor
-govendor sync
+go get ./...
```
beevee: Also this.

Makefile
litleleprikon: Removed prepare phase completely

README.md
litleleprikon: Readme was rewritten to fit go modules practices

PR #436, opened 2019-10-02
Feature/migrate to golangci lint

litleleprikon:
Change gometalinter to golangci-lint

### DIFF
--- a/senders/victorops/api/alert_test.go
+++ b/senders/victorops/api/alert_test.go
@@ -46,8 +46,9 @@ func TestCreateAlert(t *testing.T) {
 				// Send response to be tested
 			}))
 			defer server.Close()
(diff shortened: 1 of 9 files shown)


senders/victorops/api/alert_test.go
```diff
@@ -46,8 +46,8 @@ func TestCreateAlert(t *testing.T) {
 				// Send response to be tested
 			}))
 			defer server.Close()
-			client := NewClient(server.URL, server.Client())
-			err := client.CreateAlert("key", CreateAlertRequest{MessageType: Critical})
+			newClient := NewClient(server.URL, server.Client())
```
beevee: Why new client?

litleleprikon: Added TODO for investigation

PR #438, opened 2019-10-04
Add support of magic "-1" metric timestamp

### LINKED ISSUE #426: Triggers do not record a value if sending graphite "-1" magic value for timestamp
- Steps to replicate
	- Create a new trigger
	- send data using `-1` as datestamp e.g. `echo "DevOps.my_server.hdd.freespace_mbytes 9 -1" | nc localhost 2003`
	- Trigger records data sent, but no value is present
	image

Any subsequent data sent, is ignored and you have to delete the trigger and recreate it.

titusjaka:
For more details see https://graphite.readthedocs.io/en/latest/feeding-carbon.html#step-3-understanding-the-graphite-message-format

Fixes #426

### DIFF
--- a/filter/metrics_parser_test.go
+++ b/filter/metrics_parser_test.go
@@ -4,6 +4,7 @@ import (
 	"math/rand"
 	"strconv"
 	"testing"
+	"time"
 
 	. "github.com/smartystreets/goconvey/convey"
 )
@@ -74,6 +75,26 @@ func TestParseMetric(t *testing.T) {
 			So(parsedMetric.Value, ShouldEqual, validMetric.value)
 			So(parsedMetric.Timestamp, ShouldEqual, validMetric.timestamp)
 		}
+
+		Convey("Check metrics with magic '-1' timestamp", func() {
+			timeStart := time.Now().Unix()
+			magicTimestampMetrics := []ValidMetricCase{
+				{"One.two.three 123 -1", "One.two.three", "One.two.three", map[string]string{}, 123, timeStart},
+				{"One.two.three 1.23e2 -1", "One.two.three", "One.two.three", map[string]string{}, 123, timeStart},
+				{"One.two.three -123 -1", "One.two.three", "One.two.three", map[string]string{}, -123, timeStart},
+			}
+
+			for _, magicMetric := range magicTimestampMetrics {
+				parsedMetric, err := ParseMetric([]byte(magicMetric.input))
+				So(err, ShouldBeEmpty)
+				So(parsedMetric.Metric, ShouldEqual, magicMetric.metric)
+				So(parsedMetric.Name, ShouldEqual, magicMetric.name)
+				So(parsedMetric.Labels, ShouldResemble, magicMetric.labels)
+				So(parsedMetric.Value, ShouldEqual, magicMetric.value)
+				// I add 5 seconds to avoid false failures
+				So(parsedMetric.Timestamp, ShouldBeBetweenOrEqual, magicMetric.timestamp, magicMetric.timestamp+5)
+			}
+		})
 	})
 
 	Convey("Given valid metric strings with float64 timestamp, should return parsed values", t, func() {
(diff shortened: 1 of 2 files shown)


filter/metrics_parser_test.go
```diff
@@ -74,6 +75,26 @@ func TestParseMetric(t *testing.T) {
 			So(parsedMetric.Value, ShouldEqual, validMetric.value)
 			So(parsedMetric.Timestamp, ShouldEqual, validMetric.timestamp)
 		}
+
+		Convey("Check metrics with magic '-1' timestamp", func() {
+			timeStart := time.Now().Unix()
+			magicTimestampMetrics := []ValidMetricCase{
+				{"One.two.three 123 -1", "One.two.three", "One.two.three", map[string]string{}, 123, timeStart},
```
litleleprikon: Hello @titusjaka! Thank you for your contribution to moira. As I can see you you are writing tests in the style of this whole file. However I do not think this way of writing tests is good enough for reading. Let's change your tests based on this suggestions:

 * It's better to declare structures with field names. This will increase readability of test table. As for now to understand what is defined in this row I need to scroll up and check the declaration of `ValidMetricCase`.
 * The name of test case is not descriptive enough and do not give me any idea what is checked in this test and what arguments are passed to test based only on the test name. Test names also are equal between different test cases and do not allow me to define which exact test case is failed in the test run.

Based on above I will recommend you to change test cases in following style:

```suggest
{
    input: "One.two.three 123 -1",
    matric: "One.two.three",
    name: "Metric with integer value and magic -1 as time value",
    labels: map[string]string{},
    value: 123,
    timestamp: timeStart
}
```

Thank You!

filter/metrics_parser_test.go
```diff
@@ -74,6 +75,26 @@ func TestParseMetric(t *testing.T) {
...
+				{"One.two.three 1.23e2 -1", "One.two.three", "One.two.three", map[string]string{}, 123, timeStart},
+				{"One.two.three -123 -1", "One.two.three", "One.two.three", map[string]string{}, -123, timeStart},
+			}
+
+			for _, magicMetric := range magicTimestampMetrics {
+				parsedMetric, err := ParseMetric([]byte(magicMetric.input))
```
litleleprikon: I would recommend you to wrap each test case in different `Convey` call with valid name to be able to differentiate test cases in tests run report.

filter/metrics_parser_test.go
titusjaka: Hello @litleleprikon,

In my honest opinion, I would migrate all tests in Moira to [testify](https://github.com/stretchr/testify) using [table driven tests](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) instead of Convey. I assume testify is much more go-ideomatic than convey.

But I'm convinced, the test refactoring is a separate task and must not be solved in this PR. It's better to open multiple issues for the tests migration from convey to testify for each go-package, just like [Mattermost](https://github.com/mattermost/mattermost-server/issues?utf8=%E2%9C%93&q=is%3Aissue+sort%3Aupdated-desc++testify) did.

Also, it could draw additional attention of new maintainers during Hacktoberfest.

litleleprikon: @titusjaka I do not mean the refactoring of tests in the whole file. I mean that tests you added can be improved to provide better readability.

Migrating to another tests library is a huge work that should be discussed in a related issue and during this discussion I like to not to add the code which is hard to read based on the style of the rest file.

In conclusion I want to summarise my point of view: I would like to see increased quality and readability of code in every PR even if this code will be rewritten during refactoring in future.

litleleprikon: @titusjaka I do not mean the refactoring of tests in the whole file. I mean that tests you added can be improved to provide better readability.

Migrating to another tests library is a huge work that should be discussed in a related issue and during this discussion I like to not to add the code which is hard to read based on the style of the rest file.

In conclusion I want to summarise my point of view: I would like to see increased quality and readability of code in every PR even if this code will be rewritten during refactoring in future.

PR #439, opened 2019-10-09
Upgrade golangci-lint to v1.21.0

titusjaka:
It must resolve memory issues in Travis CI

https://github.com/golangci/golangci-lint/releases/tag/v1.20.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ services:
   - redis-server
 language: go
 go:
-  - 1.13.x
+  - 1.13.1
 git:
   depth: 1000
 addons:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ services:
   - redis-server
 language: go
 go:
-  - 1.13.1
+  - 1.13.x
 git:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -4,7 +4,7 @@ services:
   - redis-server
 language: go
 go:
-  - 1.13.1
+  - 1.13.x
```
litleleprikon: I would not like to add this change as I want to upgrade go versions manually and keep CI builds reproducible. And this version can change separately from Dockerfiles.

titusjaka:
> Linting is failing in travis with OOM. Seems that new version of GolancCI-Lint did not fix this. I Assume that we need to set GOGC variable again, but probably with higher value.

I set it to 50% and now it works well.

beevee:
Each next version is getting ~10s slower :-)

beevee:
Looks like this timing varies from build to build. Sometimes it's 160 seconds, sometimes 175. I'm going to merge this for the sake of more correct way of downloading golangci-lint, and removing deprecated config param.

PR #443, opened 2019-10-15
Add subscription transfer and contact deletion to CLI

### LINKED ISSUE #424: Add subscription transfer and contact deletion to CLI
Two features are often required in Moira (e.g. for ex-employees):

1. Transfer subscriptions and contacts from one user to another.
2. Delete all contacts and subscriptions for a user.

I propose we implement these features in our CLI tool.

Nixolay:
Close #424

### DIFF
--- a/cmd/cli/main.go
+++ b/cmd/cli/main.go
@@ -41,6 +41,12 @@ var (
 	plotting = flag.Bool("plotting", false, "enable images in all notifications")
 )
 
+var (
+	userDel  = flag.String("user-del", "", "Delete all contacts and subscriptions for a user")
+	fromUser = flag.String("from-user", "", "Transfer subscriptions and contacts from user.")
+	toUser   = flag.String("to-user", "", "Transfer subscriptions and contacts to user.")
+)
+
 func main() {
 	logger, dataBase := initApp()
 
@@ -69,6 +75,17 @@ func main() {
 	if *plotting {
 		if err := enablePlottingInAllSubscriptions(logger, dataBase); err != nil {
 			logger.Errorf("Failed to enable images in all notifications")
+		}
+	}
+	if *fromUser != "" || *toUser != "" {
+		if err := transferUserSubscriptionsAndContacts(dataBase, *fromUser, *toUser); err != nil {
+			logger.Error(err)
+		}
+	}
+
+	if *userDel != "" {
+		if err := deleteUser(dataBase, *userDel); err != nil {
+			logger.Error(err)
 		}
 	}
 }
--- a/cmd/cli/user.go
+++ b/cmd/cli/user.go
@@ -1,0 +1,58 @@
+package main
+
+import (
+	"github.com/moira-alert/moira"
+)
+
+func transferUserSubscriptionsAndContacts(database moira.Database, from, to string) error {
+	contactIDs, err := database.GetUserContactIDs(from)
+	if err != nil {
+		return err
+	}
+	contacts, err := database.GetContacts(contactIDs)
+	if err != nil {
+		return err
+	}
+	for _, contact := range contacts {
+		contact.User = to
+		if err = database.SaveContact(contact); err != nil {
+			return err
+		}
+	}
+
+	subscriptionIDs, err := database.GetUserSubscriptionIDs(from)
+	if err != nil {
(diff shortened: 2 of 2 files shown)


cmd/cli/user.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return err
+	}
+	for _, subscription := range subscriptions {
+		subscription.User = *toUser
+	}
+	return database.SaveSubscriptions(subscriptions)
+}
+
+func deleteUser(database moira.Database) error {
```
litleleprikon: Same comment as earlier. I will be more readable if we will receive user id here via function arguments.

cmd/cli/main.go
```diff
@@ -71,6 +77,14 @@ func main() {
 			logger.Errorf("Failed to enable images in all notifications")
 		}
 	}
+
+	if err := transferUserSubscriptionsAndContacts(dataBase); err != nil {
```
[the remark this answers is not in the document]
Nixolay: corrected

cmd/cli/main.go
```diff
@@ -71,6 +77,14 @@ func main() {
...
+		logger.Error(err)
+	}
+
+	if err := deleteUser(dataBase); err != nil {
```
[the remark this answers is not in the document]
Nixolay: corrected

cmd/cli/user.go
```diff
@@ -0,0 +1,66 @@
+package main
+
+import (
+	"github.com/moira-alert/moira"
+)
+
+func transferUserSubscriptionsAndContacts(database moira.Database) error {
+	if *fromUser == "" && *toUser == "" {
```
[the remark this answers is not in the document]
Nixolay: corrected

cmd/cli/user.go
```diff
@@ -0,0 +1,58 @@
...
+		return err
+	}
+	for _, contact := range contacts {
+		contact.User = to
+		if err = database.SaveContact(contact); err != nil {
+			return err
+		}
+	}
+
+	subscriptionIDs, err := database.GetUserSubscriptionIDs(*userDel)
```
litleleprikon: Ok we decided that passing arguments as `fromUser` and `toUser` will be more clear instead of using global variables. But what should we do with `userDel`? I think that this should be another one argument for function.

Nixolay: corrected

cmd/cli/main.go
```diff
@@ -71,6 +77,17 @@ func main() {
 			logger.Errorf("Failed to enable images in all notifications")
 		}
 	}
+	if *fromUser == "" && *toUser == "" {
+		if err := transferUserSubscriptionsAndContacts(dataBase, *fromUser, *toUser); err != nil {
+			logger.Error(err)
+		}
+	}
+
+	if *userDel == "" {
```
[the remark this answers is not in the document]
Nixolay: corrected

cmd/cli/main.go
```diff
@@ -71,6 +77,17 @@ func main() {
 			logger.Errorf("Failed to enable images in all notifications")
 		}
 	}
+	if *fromUser != "" || *toUser != "" {
```
litleleprikon: Still incorrect expression as it will be true even if only one of options will be specified `formUser` or `toUser`. In our case we need to check that `fromUser` **and** `toUser` are specified

cmd/cli/main.go
```diff
@@ -71,6 +77,17 @@ func main() {
 			logger.Errorf("Failed to enable images in all notifications")
 		}
 	}
+	if *fromUser == "" && *toUser == "" {
```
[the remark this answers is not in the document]
Nixolay: corrected

cmd/cli/main.go
litleleprikon: @beevee Am I Right in this assumptions?

Nixolay: @beevee There are two options, we make a trusted zone and they do what they need or are forbidden to do stupid things (that is, only from a normal user to a normal user)

beevee: We can transfer to/from anonymous user. Anonymous user is defined by empty string.

PR #450, opened 2019-10-30
Add twimlets support for Twilio sender

prizov:
Use TwiML generator supported by TwilioLabs [link](https://www.twilio.com/labs/twimlets/echo). In this case instead of using `voiceurl` Echo url will be provided to Twilio as callback.
Managed by `twimlets_echo` parameter in Twilio Voice sender section

### DIFF
--- a/senders/twilio/voice.go
+++ b/senders/twilio/voice.go
@@ -8,10 +8,13 @@ import (
 	"github.com/moira-alert/moira"
 )
 
+const twimletsEchoURL = "https://twimlets.com/echo?Twiml="
+
 type twilioSenderVoice struct {
 	twilioSender
 	voiceURL      string
 	appendMessage bool
+	twimletsEcho  bool
 }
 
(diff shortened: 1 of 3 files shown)


senders/twilio/voice.go
```diff
@@ -25,9 +28,14 @@ func (sender *twilioSenderVoice) SendEvents(events moira.NotificationEvents, con
 }
 
 func (sender *twilioSenderVoice) buildVoiceURL(trigger moira.TriggerData) string {
```
kamaev: Maybe it would be enough to rewrite twilio/buildVoiceURL() into something similar to webhook/buildRequestURL()
https://github.com/moira-alert/moira/blob/d796c99ea6a4ae2200ca539b1116fbad7f1b534b/senders/webhook/request.go#L52
to use "voiceurl" like here:
https://github.com/moira-alert/doc/blob/e7145befc6afcceeccf24fad3148163ba018eb39/notifier.example.yml#L74

litleleprikon: In my opinion the way URL is built here is ok. I do not thing that base URL will be changed frequently.

PR #451, opened 2019-11-05
Disabling and deleting notifications and contacts for fired users

Nixolay:
* Getting a list of users through stdin
* Add a whitelist of users and the delete flag (notifications and contacts for the user)


cmd/cli/user.go
```diff
@@ -49,10 +59,79 @@ func deleteUser(database moira.Database, user string) error {
...
+
+	if err := reader.Decode(&users); err != nil {
+		return err
+	}
+
+	return usersCleanup(database, users, conf)
+}
+
+func usersCleanup(database moira.Database, users []string, config cleanupConfig) error {
+	whitelist := make(map[string]bool)
```
litleleprikon: It is necessary to check that `config.Whitelist` is not nil. Here is PoC for bug https://play.golang.org/p/nIxb2jNdoZw

cmd/cli/user_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/moira-alert/moira/database/redis"
+	"github.com/moira-alert/moira/logging/go-logging"
+
+	"github.com/moira-alert/moira"
+
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+func TestUpdateUsers(t *testing.T) {
+	conf := getDefault()
```
litleleprikon: As I can see this test is an integration test and requires connection to Redis. I propose to skip tests like this in short mode.

```go
    if testing.Short() {
        t.Skip("skipping test in short mode.")
    }
```

cmd/cli/user.go
Nixolay: I don’t think we should think about this. We are waiting for a list, not a string line.
```data2 := []byte("hello: world")```

File: cmd/cli/user_test.go (hunk 2 above)
Comment (Nixolay) [reply cmd/cli/user_test.go]: These are not integration tests.

File: cmd/cli/user.go (hunk 1 above)
Comment (litleleprikon) [reply cmd/cli/user.go]: This string represents yaml with invalid structure and missing whitelist field. This issue is still need fix

File: cmd/cli/user_test.go (hunk 2 above)
Comment (litleleprikon) [reply cmd/cli/user_test.go]: Still this tests are not unit tests and require additional dependencies and should be skipped if I run tests in short mode.

File: cmd/cli/user.go [hunk 3]
```diff
@@ -49,10 +59,79 @@ func deleteUser(database moira.Database, user string) error {
...
+	for _, user := range config.Whitelist {
+		whitelist[user] = true
+	}
+
+	usersMap := make(map[string]bool, len(users))
```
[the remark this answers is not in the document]
Nixolay: fixed

Makefile
```diff
@@ -133,21 +133,21 @@ packages: clean build tar rpm deb
 
 .PHONY: docker_feature_images
 docker_feature_images:
-	for service in "filter" "notifier" "api" "checker" ; do \
+	for service in "filter" "notifier" "api" "checker" "cli" ; do \
```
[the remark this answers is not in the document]
Nixolay: fixed

cmd/cli/config.go
```diff
@@ -19,5 +25,8 @@ func getDefault() config {
 			Port:            "6379",
 			ConnectionLimit: 512,
 		},
+		Cleanup: cleanupConfig{
+			Whitelist: []string{""},
```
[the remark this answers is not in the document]
Nixolay: added new attr to the configuration file for handle anonymous

cmd/cli/user.go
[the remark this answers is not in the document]
Nixolay: fixed

cmd/cli/user_test.go
Nixolay: added

cmd/cli/user.go
```diff
@@ -49,10 +59,81 @@ func deleteUser(database moira.Database, user string) error {
...
+		return err
+	}
+
+	subscriptions, err := database.GetSubscriptions(subscriptionIDs)
+	if err != nil {
+		return err
+	}
+
+	for _, subscription := range subscriptions {
+		subscription.Enabled = false
```
kamaev: Unfortunately, sometimes it's just nil pointer

kamaev: BTW: i think we can also flush already scheduled notifications here (https://github.com/moira-alert/moira/blob/3fa28a373f082c49a0c1dc6b6d512470d35bb496/database/redis/notification.go#L54)

cmd/cli/user.go
```diff
@@ -49,10 +59,81 @@ func deleteUser(database moira.Database, user string) error {
...
+
+	for _, contact := range contacts {
+		if !usersMap[contact.User] {
+			usersNotFound = append(usersNotFound, contact.User)
+		}
+	}
+
+	for _, user := range usersNotFound {
+		if config.Delete {
+			if err = deleteUser(database, user); err != nil {
```
kamaev: It's a bit confusing that we have things like ("disable and save user subscritions") going right after ("delete user")
even if it would never happen
can we make this part of code more clear?

kamaev:
Is it really necessay to have this logic outside the api?
Wouldn't it be easier to use one of the hidden pages?

cmd/cli/user.go
```diff
@@ -24,13 +30,16 @@ func transferUserSubscriptionsAndContacts(database moira.Database, from, to stri
...
 		return err
 	}
+
 	subscriptions, err := database.GetSubscriptions(subscriptionIDs)
 	if err != nil {
 		return err
 	}
+
 	for _, subscription := range subscriptions {
 		subscription.User = to
```
kamaev: nil pointer again

Nixolay: I am not sure if this is a problem. Offering to exit the program? Since it will not cause a particular problem, I decided to leave it, except for a cherished processor time.
https://play.golang.org/p/5chouKdtBTU

kamaev: what type of data does GetSubscriptions return?
why do you check it's length now?

cmd/cli/user.go
```diff
@@ -49,10 +79,88 @@ func deleteUser(database moira.Database, user string) error {
...
+		return err
+	}
+
+	subscriptions, err := database.GetSubscriptions(subscriptionIDs)
+	if err != nil {
+		return err
+	}
+
+	for _, subscription := range subscriptions {
+		database.RemoveNotification(subscription.ID)
```
kamaev: it's actually "timestamp + contactID + subID"

PR #452, opened 2019-11-21
Fix nil pointer dereference in notifier

### LINKED ISSUE #437: Notifier crashes with nil pointer dereference
```
| panic: runtime error: invalid memory address or nil pointer dereference                                                                                                                   │
│ [signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xebd19a]                                                                                                                   │
│                                                                                                                                                                                           │
│ goroutine 79 [running]:                                                                                                                                                                   │
│ github.com/moira-alert/moira/notifier.fetchAvailableSeries(0x13c5660, 0xc000031030, 0xc000820000, 0xe7, 0x5d8f2584, 0x5d8f2ca3, 0x34, 0xc0005ec5f0, 0xc00024ff40, 0x1, ...)               │
│     /go/src/github.com/moira-alert/moira/notifier/plotting.go:139 +0xba                                                                                                                   │
│ github.com/moira-alert/moira/notifier.(*StandardNotifier).evaluateTriggerMetrics(0xc0003f00f0, 0x5d8f2584, 0x5d8f2ca3, 0xc000baffb0, 0x24, 0x34, 0xc000bb6400, 0x1fd, 0xc0009da800, 0x1,  │
│ ...)                                                                                                                                                                                      │
│     /go/src/github.com/moira-alert/moira/notifier/plotting.go:119 +0x255                                                                                                                  │
│ github.com/moira-alert/moira/notifier.(*StandardNotifier).buildNotificationPackagePlot(0xc0003f00f0, 0xc0003f0000, 0x1, 0x2, 0xc000baffb0, 0x24, 0xc000039140, 0x34, 0xc000bb6400, 0x1fd, │
│  ...)                                                                                                                                                                                     │
│     /go/src/github.com/moira-alert/moira/notifier/plotting.go:51 +0x36e                                                                                                                   │
│ github.com/moira-alert/moira/notifier.(*StandardNotifier).runSender(0xc0003f00f0, 0x13c5960, 0xc0000c2a00, 0xc000314420)                                                                  │
│     /go/src/github.com/moira-alert/moira/notifier/notifier.go:153 +0x157                                                                                                                  │
│ created by github.com/moira-alert/moira/notifier.(*StandardNotifier).runSenders                                                                                                           │
│     /go/src/github.com/moira-alert/moira/notifier/registrator.go:114 +0x83    ```

### PR DESCRIPTION
**Pull request by litleleprikon**
At notifier/plotting.go inside fetchAvailableSeries in case of error
returned from `metricsSource.Fetch` only
`local.ErrEvaluateTargetFailedWithPanic` was checked. Another errors did
not handled properly and function was trying to call mehods of fetch
result which became nil. That caused Nil Pointer Dereference panic. Add
proper error handling to prevent nil pointer dereference.

Closes #437

### DIFF
--- a/notifier/plotting_test.go
+++ b/notifier/plotting_test.go
@@ -1,15 +1,18 @@
 package notifier
 
 import (
+	"errors"
 	"fmt"
 	"testing"
 	"time"
 
-	"github.com/moira-alert/moira/metric_source"
+	"github.com/golang/mock/gomock"
+	"github.com/moira-alert/moira"
+	metricSource "github.com/moira-alert/moira/metric_source"
+	"github.com/moira-alert/moira/metric_source/local"
+	mockMetricSource "github.com/moira-alert/moira/mock/metric_source"
 	"github.com/op/go-logging"
 	. "github.com/smartystreets/goconvey/convey"
-
-	"github.com/moira-alert/moira"
 )
 
 func TestResolveMetricsWindow(t *testing.T) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: notifier/plotting_test.go [hunk 1]
```diff
@@ -127,3 +130,96 @@ func TestGetMetricDataToShow(t *testing.T) {
 		So(len(metricsData), ShouldEqual, len(givenSeries))
 	})
 }
+
+type MockCfg func(*mockMetricSource.MockMetricSource, *mockMetricSource.MockFetchResult) (*mockMetricSource.MockMetricSource, *mockMetricSource.MockFetchResult)
+
+func Test_fetchAvailableSeries(t *testing.T) {
```
beevee: You decided not to use Goconvey, even though all other tests in this file are Goconvey tests. Why?

litleleprikon: @beevee I believe that we decided to decrease usage of convey and finally remove convey at all, aren't we? I used the standard testing library to not increase the usage of convey. If you think that we don't need to refuse convey I can change this tests to use convey.

beevee: I remember something like this:

1. @titusjaka casually said that Goconvey sucks, and we should get rid of it. Everybody took it as an order, without much thought.
2. Later we discovered that magic feature of independently-run nested `Convey` clauses, and decided that it was interesting, and probably some of our tests rely on this behavior.

We didn't have any serious discussion after that. At least, I don't remember coming to a decision.

Here's what I think:

1. Goconvey Web UI as a useless fad, and I can easily live without it.
2. Nested `Convey` clauses that run independently is a powerful feature, even if we don't use it to the full power in most of cases right now.
3. A good-quality assertion library is a must. It makes test code so much more readable.

I don't think we have a clear plan right now. We don't have a replacement assertion library. Replacing our old tests this way, PR by PR, will take indefinite amount of time. As I see it, in a year, we'll just have 95% tests written in Goconvey, and 5% plain Go tests. Which is not a disaster, of course. Just a mess.

litleleprikon: I cannot disagree with your arguments but I have issues that I discovered during development:

 * Some of our tests define variables outside of first convey call and this variables became the shared state between "independent runs" of convey test cases. That behaviour is predictable but can lead to misunderstand of test flow. This behaviour is also can be hard for new contributors that have experience with default testing in Go. You can check the example here https://play.golang.org/p/1XtHF_QT8IW

 * Convey in my opinion is not so convenient to use with table-driven testing and provides increased testing time because for every test case convey rerun all code before the test.

 * Asserts of convey cannot be used outside of convey context that force us to use convey even if we don't need independent runs of test cases. https://play.golang.org/p/QEz3HfBbv2A

As for now I do not see the easy way to remove convey from Moira. If we will decide to refuse the convey we can try to use convey less in our tests during development and remove convey runs when we edit existing tests but in any way the transition will take a lot of time.

beevee: Goconvey is not the only assertion library on the block. We can find another one. Testify has assertions: https://godoc.org/github.com/stretchr/testify/assert

litleleprikon: Yep I know about assertions in testify. My comment was about the inconvenient usage of convey asserts. I mean that even if I want to write the test without convey I have to use convey to be able to write assertions using `So`.

litleleprikon: Changed tests to use Convey

notifier/plotting_test.go
```diff
@@ -127,3 +130,57 @@ func TestGetMetricDataToShow(t *testing.T) {
 		So(len(metricsData), ShouldEqual, len(givenSeries))
 	})
 }
+
+func Test_fetchAvailableSeries(t *testing.T) {
```
beevee: Naming is inconsistent with other tests in this file. `TestFetchAvailableSeries` would be better.

PR #460, opened 2019-12-03
Fix sending subscriptions with nil pointers elements

Nixolay:
Merge in feature/424-off-notifications-by-list-staff

### DIFF
--- a/database/redis/subscription.go
+++ b/database/redis/subscription.go
@@ -41,15 +41,21 @@ func (connector *DbConnector) GetSubscriptions(subscriptionIDs []string) ([]*moi
 	if err != nil {
 		return nil, fmt.Errorf("failed to EXEC: %s", err.Error())
 	}
+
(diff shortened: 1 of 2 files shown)


litleleprikon:
I cannot name this PR a valuable fix. I have couple of reasons to not to accept this PR.

1. This filtering hides the existence of nils in our db
2. This function accepts a list of ids and returns the list of subscriptions and I assert that this lists will have equal len and one-to-one relation between ids and subscriptions
3. You are fixing nils only in one method of db that introduce inconsistency between behaviour of different db methods. Some of methods can return nils and this method return the collection without nils. That can lead us to unexpected errors because of forgotten nil check.

@beevee @borovskyav what do you think about this PR?

borovskyav:
Согласен с Эмилем по всем трем пунктам. Логика зафиксирована по всем методам в database (даже где-то есть ишью - зафиксировать ее понятным комментом), ее ломать не нужно

PR #474, opened 2019-12-24
Add Prometheus endpoint for internal metrics

Pliner:
Fixes #380

- [x] Implement prometheus registry
- [x] Allow to collect/send metrics from prometheus/to graphite seamlessly through metrics facade(`Registry`)
- [x] Add `/metrics`endpoint to api, checker, filter, notifier
- [x] Setup prometheus in docker-compose to allow to test metrics capturing locally

PR #477, opened 2019-12-27
Metrics facade

Pliner:
Introduce Registry interface to switch go-metrics implementation(with export to graphite) to other one.

This must be done before #474.


metrics/notifier.go
```diff
@@ -0,0 +1,31 @@
...
+		EventsReceived:         registry.NewMeter(metricNameWithPrefix(prefix, "events.received")),
+		EventsMalformed:        registry.NewMeter(metricNameWithPrefix(prefix, "events.malformed")),
+		EventsProcessingFailed: registry.NewMeter(metricNameWithPrefix(prefix, "events.failed")),
+		SendingFailed:          registry.NewMeter(metricNameWithPrefix(prefix, "sending.failed")),
+		SendersOkMetrics:       registry.NewMetersCollection(),
+		SendersFailedMetrics:   registry.NewMetersCollection(),
+	}
+}
+
+func metricNameWithPrefix(prefix, metric string) string {
```
litleleprikon: I think it should be in another file. This function is used in different files like `checker.go` and `notifier.go`. Probably better wil be to move it to `graphite.go`

Pliner: image

PR #487, opened 2020-01-23
Add templates to trigger description #484

### LINKED ISSUE #484: Add templates to trigger description
It's often useful to add a Grafana dashboard link to a trigger description. But sometimes you have a templated dashboard. That is, you need to pass a specific variable value (e.g. host name) to open a dashboard that is relevant to the alert.

For example:
https://grafana.yourhost.com/some-dashboard?var-host=-METRICNAME-

We can add templating capabilities to our trigger descriptions to fill and render values at notification generation time.

Some things to consider:

1. What if a notification contains several events? How do we render the description? Do we pass an array of values to the template?
2. What do we render in web UI? A template? This can make links on the web unusable.
3. Which template language do we use? Which variables do we pass when rendering the template? Which templating capabilities do we allow so we don't compromise security?

Nixolay:
* Add templates to trigger description
* Add presentation of trigger description in non-editing for API

Close #484

### DIFF
--- a/api/handler/trigger.go
+++ b/api/handler/trigger.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/go-chi/chi"
 	"github.com/go-chi/render"
+	"github.com/moira-alert/moira"
 	"github.com/moira-alert/moira/metric_source/local"
 	"github.com/moira-alert/moira/metric_source/remote"
 
@@ -20,7 +21,8 @@ import (
 func trigger(router chi.Router) {
 	router.Use(middleware.TriggerContext)
 	router.Put("/", updateTrigger)
-	router.Get("/", getTrigger)
+	router.With(middleware.TriggerContext,
+		middleware.Populate(false)).Get("/", getTrigger)
 	router.Delete("/", removeTrigger)
 	router.Get("/state", getTriggerState)
 	router.Route("/throttling", func(router chi.Router) {
@@ -35,6 +37,7 @@ func trigger(router chi.Router) {
 func updateTrigger(writer http.ResponseWriter, request *http.Request) {
 	triggerID := middleware.GetTriggerID(request)
 	trigger := &dto.Trigger{}
+
 	if err := render.Bind(request, trigger); err != nil {
 		switch err := err.(type) {
 		case local.ErrParseExpr, local.ErrEvalExpr, local.ErrUnknownFunction:
@@ -50,7 +53,17 @@ func updateTrigger(writer http.ResponseWriter, request *http.Request) {
 		default:
 			render.Render(writer, request, api.ErrorInternalServer(err))
 		}
+
 		return
+	}
+
+	if trigger.Desc != nil {
+		triggerData := moira.TriggerData{Desc: *trigger.Desc, Name: trigger.Name}
+		if _, err := triggerData.GetPopulatedDescription(moira.NotificationEvents{}); err != nil {
+			render.Render(writer, request, api.ErrorRender(
+				fmt.Errorf("You have an error in your Go template: %v", err)))
+			return
+		}
 	}
 
 	timeSeriesNames := middleware.GetTimeSeriesNames(request)
@@ -84,6 +97,18 @@ func getTrigger(writer http.ResponseWriter, request *http.Request) {
 		render.Render(writer, request, err)
 		return
 	}
+
+	if needToPopulate := middleware.GetPopulated(request); needToPopulate && trigger.Desc != nil {
+		triggerData := moira.TriggerData{Desc: *trigger.Desc, Name: trigger.Name}
+
+		eventsList, err := controller.GetTriggerEvents(database, triggerID, 0, 3)
+		if err != nil {
+			render.Render(writer, request, err)
+		}
+
+		*trigger.Desc, _ = triggerData.GetPopulatedDescription(eventsList.List)
+	}
+
 	if err := render.Render(writer, request, trigger); err != nil {
 		render.Render(writer, request, api.ErrorRender(err))
 	}
--- a/api/middleware/middleware.go
+++ b/api/middleware/middleware.go
@@ -19,6 +19,7 @@ var (
 	databaseKey          ContextKey = "database"
 	searcherKey          ContextKey = "searcher"
 	triggerIDKey         ContextKey = "triggerID"
+	populateKey          ContextKey = "populated"
 	contactIDKey         ContextKey = "contactID"
 	tagKey               ContextKey = "tag"
 	subscriptionIDKey    ContextKey = "subscriptionID"
@@ -46,6 +47,11 @@ func GetLogin(request *http.Request) string {
 // GetTriggerID gets TriggerID string from request context, which was sets in TriggerContext middleware
 func GetTriggerID(request *http.Request) string {
 	return request.Context().Value(triggerIDKey).(string)
+}
+
+// GetPopulated get populate bool from request context, which was sets in TriggerContext middleware
+func GetPopulated(request *http.Request) bool {
+	return request.Context().Value(populateKey).(bool)
 }
 
 // GetTag gets tag string from request context, which was sets in TagContext middleware
--- a/datatypes.go
+++ b/datatypes.go
@@ -3,6 +3,7 @@ package moira
 import (
 	"bytes"
 	"fmt"
+	"html/template"
 	"math"
 	"strconv"
 	"strings"
@@ -103,6 +104,23 @@ func (event *NotificationEvent) CreateMessage(location *time.Location) string {
 // NotificationEvents represents slice of NotificationEvent
 type NotificationEvents []NotificationEvent
 
+type templateData struct {
+	Trigger templateTrigger
+	Events  []templateEvent
+}
+
+type templateEvent struct {
+	Metric         string
+	MetricElements []string
+	Timestamp      int64
+	Value          *float64
+	State          State
+}
+
+type templateTrigger struct {
+	Name string `json:"name"`
+}
+
 // TriggerData represents trigger object
 type TriggerData struct {
 	ID         string   `json:"id"`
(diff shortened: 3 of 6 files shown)


datatypes.go
```diff
@@ -115,6 +128,19 @@ type TriggerData struct {
 	Tags       []string `json:"__notifier_trigger_tags"`
 }
 
+func (trigger *TriggerData) TemplateDescription(data map[string]interface{}) error {
```
litleleprikon: Name `TemplateDescription` do not explain what this method do. This method fills the trigger description template with following data. Probably `PopulateDescription` will be better choice.

litleleprikon: I do not like the way of passing data as a map. I would like to see a struct that contains `TemplateEvent` and `TemplateTrigger` structs as a fields and usage of this structure will be better.

datatypes.go
```diff
@@ -103,6 +104,18 @@ func (event *NotificationEvent) CreateMessage(location *time.Location) string {
 // NotificationEvents represents slice of NotificationEvent
 type NotificationEvents []NotificationEvent
 
+type TemplateEvent struct {
```
litleleprikon: The name of this struct do not explain what this struct is about. Let's rename it to `EventTemplateData`

datatypes.go
```diff
@@ -103,6 +104,18 @@ func (event *NotificationEvent) CreateMessage(location *time.Location) string {
...
+	Metric         string
+	MetricElements []string
+	Timestamp      int64
+	Value          *float64
+	State          State
+}
+
+type TemplateTrigger struct {
```
litleleprikon: Same comment about the name as for previous struct

datatypes.go
```diff
@@ -115,6 +128,19 @@ type TriggerData struct {
 	Tags       []string `json:"__notifier_trigger_tags"`
 }
 
+func (trigger *TriggerData) TemplateDescription(data map[string]interface{}) error {
+	buffer := new(bytes.Buffer)
+
+	triggerTemplate := template.Must(template.New("moira").Parse(trigger.Desc))
```
litleleprikon: In this certain case template name do not mean anything even so let's name template as a `trigger-description`

notifier/notifier.go
```diff
@@ -160,6 +161,9 @@ func (notifier *StandardNotifier) runSender(sender moira.Sender, ch chan Notific
 				notifier.logger.Errorf(buildErr)
 			}
 		}
+
+		notifier.BuildTriggerDescription(pkg.Events, &pkg.Trigger)
```
litleleprikon: I believe that we should cover this function with tests that uses templates

datatypes_test.go
```diff
@@ -244,6 +244,61 @@ func TestTriggerData_GetTags(t *testing.T) {
...
+				"MetricElements: []\n"+
+				"Timestamp: 0\n"+
+				"Value: <nil>\n"+
+				"State: \n"+
+				"\n"+
+				"https://grafana.yourhost.com/some-dashboard?var-host=1&var-host=2\n")
+		})
+
+		Convey("Test description without templates", func() {
+			anotherText := "Anther text"
```
litleleprikon: Typo here

api/handler/trigger.go
```diff
@@ -79,11 +81,21 @@ func getTrigger(writer http.ResponseWriter, request *http.Request) {
...
 		panic("Test for multi line logs")
 	}
+
 	trigger, err := controller.GetTrigger(database, triggerID)
 	if err != nil {
 		render.Render(writer, request, err)
 		return
 	}
+
+	value, templated := request.URL.Query()["empty-templated"]
```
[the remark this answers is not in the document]
Nixolay: По сообщения я добавил не пустой и в треде по дню писал что добавил еще  с данными. Не уверен что нужно оставить с пустыми. Пока не запушил раз нормально не протетировал у себя.

datatypes.go
Nixolay: Уже поправлено будет после того как протетсирую

api/handler/trigger.go
```diff
@@ -79,11 +80,30 @@ func getTrigger(writer http.ResponseWriter, request *http.Request) {
...
 		panic("Test for multi line logs")
 	}
+
 	trigger, err := controller.GetTrigger(database, triggerID)
 	if err != nil {
 		render.Render(writer, request, err)
 		return
 	}
+
+	if _, templated := request.URL.Query()["templated"]; templated && trigger.Desc != nil {
```
litleleprikon: You still do query arguments parsing wrong. You need to implement a middleware that parses the query param. Example: https://github.com/moira-alert/moira/blob/master/api/middleware/context.go#L127-L145

Then you need to implement a function that retrieves data from context. Example: https://github.com/moira-alert/moira/blob/master/api/middleware/middleware.go#L74-L77

Then you need to register middleware in router. Example https://github.com/moira-alert/moira/blob/master/api/handler/trigger.go#L32

Remove this code from handler

litleleprikon: I was saying that `templated` is not a name for query parameter that describes it perfectly. Use for example `render_description` or something like that

datatypes.go
```diff
@@ -115,6 +128,40 @@ type TriggerData struct {
...
+			templateEvent.Value = event.Value
+		}
+
+		templateEvents = append(templateEvents, templateEvent)
+	}
+
+	data := map[string]interface{}{
+		"Trigger": triggerTemplateData{Name: trigger.Name},
+		"Events":  templateEvents,
+	}
```
litleleprikon: Please, use data structures instead of maps for templates

notifier/notifier.go
```diff
@@ -160,6 +160,9 @@ func (notifier *StandardNotifier) runSender(sender moira.Sender, ch chan Notific
 				notifier.logger.Errorf(buildErr)
 			}
 		}
+
+		pkg.Trigger.PopulateDescription(pkg.Events)
```
litleleprikon: Still do not see the tests for notifications with templates in descriptions

api/middleware/middleware.go
```diff
@@ -46,6 +47,16 @@ func GetTriggerID(request *http.Request) string {
 	return request.Context().Value(triggerIDKey).(string)
 }
 
+// GetTriggerID gets TriggerID string from request context, which was sets in TriggerContext middleware
+func GetTemplated(request *http.Request) bool {
+	// value, ok := request.URL.Query()["templated"]
+	// if ok {
+	// 	return fmt.Sprint(value) == "[true]"
+	// }
+	// return false
```
litleleprikon: Remove comments please

datatypes.go
```diff
@@ -115,6 +128,40 @@ type TriggerData struct {
 	Tags       []string `json:"__notifier_trigger_tags"`
 }
 
+func (trigger *TriggerData) PopulateDescription(events NotificationEvents) error {
```
litleleprikon: Probably better to not to edit current trigger with populated description and create a new trigger instead of it?

datatypes.go
litleleprikon: This issue is still not fixed

api/middleware/middleware.go
```diff
@@ -19,6 +19,7 @@ var (
 	databaseKey          ContextKey = "database"
 	searcherKey          ContextKey = "searcher"
 	triggerIDKey         ContextKey = "triggerID"
+	templatedKey         ContextKey = "templated"
```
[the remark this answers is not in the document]
Nixolay: mb description-by-template?

api/handler/trigger.go
```diff
@@ -79,11 +95,33 @@ func getTrigger(writer http.ResponseWriter, request *http.Request) {
...
 		render.Render(writer, request, err)
 		return
 	}
+
+	if templated := middleware.GetTemplated(request); templated && trigger.Desc != nil {
+		triggerData := moira.TriggerData{
+			ID:   trigger.ID,
+			Name: trigger.Name,
+			Desc: *trigger.Desc,
+		}
```
litleleprikon: I definitely do not like creation of new trigger here. I propose to change `PopulateDescription`  to return a new instance of trigger with a populated description instead of just string.

datatypes.go
```diff
@@ -6,6 +6,7 @@ import (
 	"math"
 	"strconv"
 	"strings"
+	"text/template"
```
litleleprikon: I think that we need to use `html/template` instead of `text/template`

Nixolay: Why?

notifier/notifier.go
```diff
@@ -160,6 +160,12 @@ func (notifier *StandardNotifier) runSender(sender moira.Sender, ch chan Notific
 				notifier.logger.Errorf(buildErr)
 			}
 		}
+
+		desc, err := pkg.Trigger.PopulateDescription(pkg.Events)
+		if err == nil {
+			pkg.Trigger.Desc = desc
```
beevee: Maybe log a warning if `err` is not `nil`? It's wrong to silently ignore malformed descriptions.

datatypes.go
beevee: Official documentation for `html/template` states that:

> Package template (html/template) implements data-driven templates for generating HTML output safe against code injection. It provides the same interface as package text/template and should be used instead of text/template whenever the output is HTML.

So we need to choose `html/template` if there is a risk of HTML injection in description field using our templating mechanism.

datatypes.go
```diff
@@ -115,6 +133,36 @@ type TriggerData struct {
...
+
+	for _, data := range events {
+		event := templateEvent{
+			Metric:         data.Metric,
+			MetricElements: strings.Split(data.Metric, "."),
+			Timestamp:      data.Timestamp,
+			State:          data.State,
+		}
+
+		if data.Value != nil {
```
beevee: Why do you need to check for `nil` here? What's going to change if you remove this check?

datatypes.go
```diff
@@ -115,6 +133,36 @@ type TriggerData struct {
 	Tags       []string `json:"__notifier_trigger_tags"`
 }
 
+func (trigger TriggerData) PopulateDescription(events NotificationEvents) (string, error) {
```
beevee: Contrary to its name, this method does not populate a trigger's description. It returns a populated description string. You should either rename this to `GetPopulatedDescription` or return a full new instance of a trigger with populated description field, not just string.

api/middleware/middleware.go
```diff
@@ -46,6 +47,11 @@ func GetTriggerID(request *http.Request) string {
 	return request.Context().Value(triggerIDKey).(string)
 }
 
+// GetTemplated get populate bool from request context, which was sets in TriggerContext middleware
```
beevee: Wrong method name in comment

notifier/notifier.go
```diff
@@ -160,6 +160,13 @@ func (notifier *StandardNotifier) runSender(sender moira.Sender, ch chan Notific
 				notifier.logger.Errorf(buildErr)
 			}
 		}
+
+		desc, err := pkg.Trigger.GetPopulatedDescription(pkg.Events)
+		if err == nil {
+			notifier.logger.Errorf("Error populate description: %v", err)
```
beevee: Sooooo, if error is nil, we log an error. Brilliant.

PR #490, opened 2020-02-11
WIP: fix incorrect parsing of multiple equal signs in label

### LINKED ISSUE #489: Incorrect parsing of graphite tags with equal signs in value
# BUG REPORT

## What did you expect to see?
Metrics that have equal signs in tag values should be accepted by Moira. For example `my.series;tag1=value1;tag2=value=2` is a valid metric with tags following carbon specification.

https://github.com/graphite-project/graphite-web/blob/7149908675664f341c91dcacfff83c55505564fa/docs/tags.rst#carbon

## What did you see instead?
Moira filters will drop this metric as an invalid with following message:

```
INFO    cannot parse input: cannot parse metric: 'my.series;tag1=value1;tag2=value=2' (too many equal-separated items: 'tag2=value=2')
```

idoqo:
This fixes #489 (incorrect parsing of multiple equal signs in label).

### DIFF
--- a/filter/metrics_parser_test.go
+++ b/filter/metrics_parser_test.go
@@ -64,6 +64,10 @@ func TestParseMetric(t *testing.T) {
 			{"One.two.three;four= 123 1234567890", "One.two.three;four=", "One.two.three", map[string]string{"four": ""}, 123, 1234567890},
 			{"One.two.three;four=five;six=seven 123 1234567890", "One.two.three;four=five;six=seven", "One.two.three", map[string]string{"four": "five", "six": "seven"}, 123, 1234567890},
 			{"One.two.three;four=five;six=seven=eight 123 1234567890", "One.two.three;four=five;six=seven=eight", "One.two.three", map[string]string{"four": "five", "six": "seven=eight"}, 123, 1234567890},
+			{"One.two.three;four=five;six=seven=eight=nine 123 1234567890", "One.two.three;four=five;six=seven=eight=nine",
+				"One.two.three", map[string]string{"four": "five", "six": "seven=eight=nine"}, 123, 1234567890},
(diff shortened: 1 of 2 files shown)


filter/metrics_parser_test.go
```diff
@@ -64,6 +63,7 @@ func TestParseMetric(t *testing.T) {
 			{"One.two.three;four=five 123 1234567890", "One.two.three;four=five", "One.two.three", map[string]string{"four": "five"}, 123, 1234567890},
 			{"One.two.three;four= 123 1234567890", "One.two.three;four=", "One.two.three", map[string]string{"four": ""}, 123, 1234567890},
 			{"One.two.three;four=five;six=seven 123 1234567890", "One.two.three;four=five;six=seven", "One.two.three", map[string]string{"four": "five", "six": "seven"}, 123, 1234567890},
+			{"One.two.three;four=five;six=seven=eight 123 1234567890", "One.two.three;four=five;six=seven=eight", "One.two.three", map[string]string{"four": "five", "six": "seven=eight"}, 123, 1234567890},
```
litleleprikon: Hi @idoqo I recommend you to add test case with at least two equal signs in value. This case will not work in your implementation.

Example:
```go
{"One.two.three;four=five;six=seven=eight=nine 123 1234567890", "One.two.three;four=five;six=seven=eight=nine", "One.two.three", map[string]string
```

litleleprikon:
@borovskyav Check the new behaviour of tags parsing

filter/metrics_parser_test.go
idoqo: Thanks a lot for the catch. I've used a for loop to resolve it and added two new test cases. I don't know if the solution is okay though.

litleleprikon:
@idoqo Thank you

PR #491, opened 2020-02-16
Shift plot window when not realtime (#188)

### LINKED ISSUE #188: Fix last two points on graphs when allowRealtimeAlerting is disabled
В графиках видна дыра в конце. Метрика чекается на минуту позже, чем должна, если она не realtime

A1bemuth:
Shift plot window one minute to the left when forcing realtime window to avoid empty space at right boundary. 
Fixes #188


A1bemuth:
Here's my naive attempt at fixing the issue. From what i've concluded, the empty dot on the right side of the plot happens due to some rounding up when fetching series and unpacking them with 60s retention. Hence, when we force the default 30m window it has to be shifted 1m to the left.

A1bemuth:
Nevermind, I can't even reproduce the bug, this fix is nonsensical.

PR #492, opened 2020-02-24
[DOC] Updated feed graphite data command

hv7214:
# PR Summary
Added `-N` flag, which closes the network socket after EOF on the input. 

**Note**
This PR doesn't bring any breaking-change or bug-fix. It's just adds user comfortability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ See more on our [documentation page](https://moira.readthedocs.io/en/latest/inst
 Feed data in Graphite format to `localhost:2003`:
 
 ```bash
-echo "local.random.diceroll 4 `date +%s`" | nc localhost 2003
+echo "local.random.diceroll 4 `date +%s`" | nc -N localhost 2003
 ```
 
 Configure triggers at `localhost:8080` using your browser.


beevee:
> This PR doesn't bring any breaking-change

Or does it? In fact, `-N` breaks compatibility with MacOS and a number of other systems. `nc` flags are not very well standardized across platforms.

I'm disappointed with the amount of effort you put in this PR. At least you could consult with the original Graphite documentation (https://graphite.readthedocs.io/en/latest/feeding-carbon.html#the-plaintext-protocol) that provides info on multiple possible flags.

hv7214:
@beevee i'm really sorry for not diving deeper. I think it's better to use `Ctrl + C` to close the socket manually and write the same in README.

beevee:
`Ctrl-C` won't work in automated scenarios when you want to write a script to send metrics.

hv7214:
I guess that is possible in shell script.

hv7214:
oh i see, `Ctrl + C` is a signal. Though we can send `Ctrl + C` signal in bash script.

hv7214:
@litleleprikon closing this PR

Issue #469, opened 2019-12-10
Report who and why turned off notifications


litleleprikon:
In Moira when notifications turned off is very hard to determine who and why turned off notifications. I propose to add a text field that should be filled with purpose when someone will switch off notifications. When notifications turned off Moira should display user who disabled notifications and purpose.

idoqo:
Hi @litleleprikon, I'd love to work on this if it's still available. I'm not so clear if the notifications here refer to Subscriptions though.

hv7214:
@idoqo are you working on this ?

idoqo:
Oh No, not yet. Wasn't sure if the notification meant turning off a trigger or deleting a subscription.

vrongmeal:
Hey! I'm new here. Can I take this up?

litleleprikon:
Hi @idoqo notifications turned off means that this switcher on `/notifications` page is switched off

<img width="225" alt="Screenshot 2020-02-21 at 11 14 54" src="https://user-images.githubusercontent.com/5058463/75016271-a7cb3d00-549b-11ea-8834-af66a257bdd8.png">

hv7214:
@litleleprikon can i pick this up ?

idoqo:
> Hi @idoqo notifications turned off means that this switcher on `/notifications` page is switched off
> 
> <img alt="Screenshot 2020-02-21 at 11 14 54" width="225" src="https://user-images.githubusercontent.com/5058463/75016271-a7cb3d00-549b-11ea-8834-af66a257bdd8.png">

Okay got it, will be on it then.

shashikant231:
Hi @litleleprikon  can you describe  some steps to reproduce this behaviour ?

imshubh4141:
@idoqo Is the issue still open?

idoqo:
Yeah it is, there's an incomplete implementation here https://github.com/moira-alert/moira/pull/520 so feel free to look at it for if you need some pointers.

PR #493, opened 2020-02-24
fix(api): Log reason for turning off notifications

idoqo:
# PR Summary
The `NotifierState` struct in `dto/health.go` allows both state and message. But when the state is ERROR and notifications are off, we currently show a default `dto.ErrorMessage`.
This change allows for custom error messages which serves as the reason for turning notifications off.

Additional information
The text field has not been added to the front-end yet, I'm hoping to do that after the changes to the API gets merged.

Relates to #469

### DIFF
--- a/api/controller/health.go
+++ b/api/controller/health.go
@@ -13,17 +13,26 @@ func GetNotifierState(database moira.Database) (*dto.NotifierState, *api.ErrorRe
 		return nil, api.ErrorInternalServer(err)
 	}
 
-	notifierState := dto.NotifierState{State: state}
-	if state == moira.SelfStateERROR {
-		notifierState.Message = dto.ErrorMessage
+	message, err  := database.GetNotifierMessage()
+	if err != nil {
+		return nil, api.ErrorInternalServer(err)
 	}
 
+	notifierState := dto.NotifierState{State: state, Message: message}
 	return &notifierState, nil
 }
 
 // UpdateNotifierState update current notifier state
 func UpdateNotifierState(database moira.Database, state *dto.NotifierState) *api.ErrorResponse {
 	err := database.SetNotifierState(state.State)
+	if err != nil {
+		return api.ErrorInternalServer(err)
+	}
+
+	if state.State == moira.SelfStateERROR && state.Message == "" {
+		state.Message = dto.ErrorMessage
+	}
+	err = database.SetNotifierMessage(state.Message)
 	if err != nil {
 		return api.ErrorInternalServer(err)
 	}
--- a/database/redis/selfstate.go
+++ b/database/redis/selfstate.go
@@ -3,6 +3,7 @@ package redis
 import (
 	"github.com/gomodule/redigo/redis"
 	"github.com/moira-alert/moira"
+	"github.com/moira-alert/moira/api/dto"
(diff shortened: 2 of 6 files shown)


database/redis/selfstate.go
```diff
@@ -3,6 +3,7 @@ package redis
 import (
 	"github.com/gomodule/redigo/redis"
 	"github.com/moira-alert/moira"
+	"github.com/moira-alert/moira/api/dto"
```
litleleprikon: I do not like usage of `dto` package in database layer. the only reason of DTO is a conversion to JSON and from JSON in API.

database/redis/selfstate.go
```diff
@@ -68,7 +69,25 @@ func (connector *DbConnector) SetNotifierState(health string) error {
...
+	c := connector.pool.Get()
+	defer c.Close()
+	message, err := redis.String(c.Do("GET", selfStateNotifierMessage))
+	if err == redis.ErrNil {
+		message = dto.DefaultMessage
+	}
+	return message, err
+}
+
+func (connector *DbConnector) SetNotifierMessage(message string) error {
```
litleleprikon: I think that notifier message should be set in a `SetNotifierState`. It will prevent a cases when notifier state is changed but notifier message is not. In this cases notifier state and message will be in inconsisten state and will increase confusion on incident investigation. Especially I can see that in this PR notification message is not changed in self state checker.

api/controller/health.go
```diff
@@ -13,11 +13,12 @@ func GetNotifierState(database moira.Database) (*dto.NotifierState, *api.ErrorRe
 		return nil, api.ErrorInternalServer(err)
 	}
 
-	notifierState := dto.NotifierState{State: state}
-	if state == moira.SelfStateERROR {
-		notifierState.Message = dto.ErrorMessage
+	message, err  := database.GetNotifierMessage()
+	if err != nil {
```
litleleprikon: Why `if state == moira.SelfStateERROR` is disappeared here?

idoqo: `notifierState.Message = dto.ErrorMessage` was setting the notifier message automatically, but now, it fetches the message from the database instead.

database/redis/selfstate.go
idoqo: Yeah that's true, I'm not sure how to go about that though. Can we modify the arguments for `SetNotifierState` in the `Database` interface so that it takes two strings (`state` and `message`)?

database/redis/selfstate.go
idoqo: It was only using a variable from the `dto` package, I've changed it to remove the import now.

database/redis/selfstate.go
litleleprikon: Yes, this was my point

litleleprikon: Even more I want to see notifier state as a tuple of two values: boolean state and string message. And I want to see that this two values are changed atomically in one place.

idoqo:
I'm creating a new PR with the things learnt from this one  as the divergence seems to quite much, so I'd go ahead and close this if that's okay.

beevee:
Sure, go ahead.

PR #494, opened 2020-02-24
Add description to message in telegram sender (moira-alert#353)

kostya05983:
# PR Summary
Add description without markdown to telegram sender.
Additional information
I think it's good feature without markdown which can help to understand alerts in team.
Closes/Relates #issue
#353

### DIFF
--- a/senders/telegram/send.go
+++ b/senders/telegram/send.go
@@ -3,7 +3,9 @@ package telegram
 import (
 	"bytes"
 	"fmt"
+	"github.com/moira-alert/moira/senders"
 	"gopkg.in/tucnak/telebot.v2"
+	"strings"
 
 	"github.com/moira-alert/moira"
 )
@@ -18,10 +20,8 @@ const (
 )
 
 const (
-	photoCaptionMaxCharacters     = 1024
-	messageMaxCharacters          = 4096
-	additionalInfoCharactersCount = 400
-	maxDescriptionSize            = 450
+	photoCaptionMaxCharacters = 1024
+	messageMaxCharacters      = 4096
 )
 
 var characterLimits = map[messageType]int{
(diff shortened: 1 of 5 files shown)


senders/telegram/send.go
```diff
@@ -50,7 +50,7 @@ func (sender *Sender) buildMessage(events moira.NotificationEvents, trigger moir
 	tags := trigger.GetTags()
 	emoji := emojiStates[state]
 
-	title := fmt.Sprintf("%s%s %s %s (%d)\n", emoji, state, trigger.Name, tags, len(events))
+	title := fmt.Sprintf("%s%s %s %s (%d)\n%s\n", emoji, state, trigger.Name, tags, len(events), trigger.Desc)
```
beevee: This is not going to work. Any description longer than ~500-600 characters will silently break Telegram notifications due to message length limit (see lines 22..24). You need to cut description according to the limits.

borovskyav: You also need to convert the description to a format that supports telegram, as is done, for example, in slack (there we use the blackfriday-slack package), there’s no such library for a telegram and we need to implement on your own

beevee: We can tolerate bad markup for now. But length limit is a show-stopper: if we merge this PR as-is, some users will silently lose their alerts.

kostya05983: @beevee I fixed it.

senders/telegram/send.go
```diff
@@ -3,7 +3,6 @@ package telegram
 import (
 	"bytes"
 	"fmt"
-
```
beevee: We sort imports with `goimports`, so this empty line should stay.

We probably need to set up a linter for this @litleleprikon.

senders/telegram/send.go
```diff
@@ -46,11 +46,18 @@ func (sender *Sender) SendEvents(events moira.NotificationEvents, contact moira.
 
 func (sender *Sender) buildMessage(events moira.NotificationEvents, trigger moira.TriggerData, throttled bool, maxChars int) string {
 	var buffer bytes.Buffer
+	var descr string
```
litleleprikon: I do not like the reduction of word description here. I do not see the strong reasons to use it instead of `description`

senders/telegram/send_test.go
```diff
@@ -34,6 +34,7 @@ func TestBuildMessage(t *testing.T) {
 			actual := sender.buildMessage([]moira.NotificationEvent{event}, trigger, false, messageMaxCharacters)
 			expected := `💣NODATA Trigger Name [tag1][tag2] (1)
 
+
```
litleleprikon: Can you add at least one test that covers a description?

senders/telegram/send.go
```diff
@@ -22,6 +21,7 @@ const (
 	photoCaptionMaxCharacters     = 1024
 	messageMaxCharacters          = 4096
 	additionalInfoCharactersCount = 400
+	maxDescriptionSize            = 450
```
beevee: But why 450?

You have actual `maxChars` value (1024 or 4096) in `buildMessage`. You have full information on `events` in `buildMessage`. Let's say `maxChars` is 4096 and you have only one event. Is it really necessary to cut description to just 450 chars?

You can refer to https://github.com/moira-alert/moira/blob/d87d8757234c5fe688d8388751fd4b53fdd4b22e/senders/slack/slack.go#L76 and reuse code to implement this better.

senders/telegram/send.go
beevee: Or just `desc`, since it's equal to `trigger.Desc`.

senders/telegram/send.go
beevee: Or it can be `desc`, since we use `trigger.Desc`.

senders/telegram/send.go
litleleprikon: We have a problems only with one linter in golangci-lint. This linter is `golint` as we have a bunch of unchecked returned errors. I believe that we can setup more strict linter. I will check it

litleleprikon:
@kostya05983 would you like to fix build so we can review your PR?

kostya05983:
@litleleprikon yes, I would like.

kostya05983:
@litleleprikon. Yes, you can rewiew this. I reuse code from slack for telegram. And I'm sorry, but I deleted emoj from telegram, I think it's not the best feature and maybe we need emoj in slack too? What do you think?

kostya05983:
@beevee I reused code as you say, can you review it now? I think, changes requested isn't worked properly as I had made commit ammend.

AleksandrMatsko:
Close, because feature implemented in: https://github.com/moira-alert/moira/pull/1075

PR #495, opened 2020-02-25
fix(notifier): Move plot boundaries

### LINKED ISSUE #188: Fix last two points on graphs when allowRealtimeAlerting is disabled

A1bemuth:
Fixed the missing point on the right side of plots.

# PR Summary
Fix the missing right dot in non-realtime triggers due to rounding in redis.
## Example:
Say we have two metric values for some non-realtime trigger.
|value|          ts          |        score       |    tsTime        |  scoreTime    |
|------|----------------|----------------|---------------|----------------|
|11    |   1582287284  | 1582287300  | 12:14:44 PM  |   12:15:00 PM|
|15    |   1582287344  | 1582287360  | 12:15:44 PM  |   12:16:00 PM|

Metric value 11 is the event that we want on a plot. 15 is a value that should get trimmed cause the trigger is not realtime. When notifier tries to fetch data, it specifies the right boundary as event's time plus one minute, so 12:15:45 PM (1582287345). Since fetch ranges values by score, we only get value 11, that later gets trimmed by graphite's expression evaluator. To fix that we need to first round the time to retention, and then increment it by one minute.

## Changes:
- Round boundaries to retention at `notifier/plotting.go` to ensure we get all the needed metric points from redis. This is required because metric's score in redis is a timestamp rounded to retention.

Closes #188

### DIFF
--- a/notifier/plotting.go
+++ b/notifier/plotting.go
@@ -17,6 +17,8 @@ const (
 	defaultTimeShift = 1 * time.Minute
 	// defaultTimeRange is default time range to fetch timeseries
 	defaultTimeRange = 30 * time.Minute
+	// defaultRetentionSeconds is the most common metric retention
+	defaultRetentionSeconds = 60
 )
 
 // errFetchAvailableSeriesFailed is used in cases when fetchAvailableSeries failed after retry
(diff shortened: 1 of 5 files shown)


notifier/plotting.go
```diff
@@ -86,22 +91,21 @@ func resolveMetricsWindow(logger moira.Logger, trigger moira.TriggerData, pkg No
...
-	return alignToMinutes(defaultFrom), defaultTo
+	return defaultFrom, defaultTo
 }
 
-func alignToMinutes(unixTime int64) int64 {
-	unixTime -= unixTime % 60
-	return unixTime
+func roundToRetention(unixTime int64) int64 {
+	return moira.RoundToNearestRetention(unixTime, defaultRetentionSeconds)
 }
```
litleleprikon: Not sure why we need to implement this simple function as we can use `moira.RoundToNearestRetention` with default retention everywhere above.

A1bemuth: I've introduced it for brevity since rounding happens like 10 times in `resolveMetricsWindow` and each time we round to same retention.

PR #499, opened 2020-03-02
refactor: moira by Standard Go Project Layout

Nixolay:
I want to bring beauty to Moira, by example:
https://github.com/golang-standards/project-layout


beevee:
image

I see multiple people displeased with these _standards_, so I'm not sure.

litleleprikon:
Ok, I am happy to see the attempts to increase the quality of our code. I am sure that we can change layout of Moira to be more consistent and observable. I want just say that changes like this should be discussed a lot and all the process should be investigated before changes. The serious issue for this kind of changes is that it will cause merge conflicts and at current time we cannot accept this issue as we have a huge amount of new contributors. I am also agree with comment above that repository you pointing to is not a standard at all and have serious issues.

I propose to close this PR and fill the issue for this change to track investigation.

Nixolay:
I read these issues and did not see anything serious problems.

Issue #254, opened 2019-02-12
[ plotting ] Empty plots are possible


kamaev:
It takes to:
- do not use buff.bytes() for NotificationPackage when Render() fails, use []byte{} instead
- [limits/resolveLimits] remove or refactor MinAndMaxOfTime to provide valid and different from/to's
- add tests for different retention intervals

borovskyav:
[metric.txt](https://github.com/moira-alert/moira/files/2857315/metric.txt)

kamaev:
Main goal is to avoid errors like **infinite {x,y}-range delta** from wcharczuk/go-chart

gane5hvarma:
Hi i would like to work on this issue.is anyone working on this?

beevee:
@gane5hvarma OK, you can take this task.

srv-twry:
@beevee I'll be happy to take this one.

beevee:
> @beevee I'll be happy to take this one.

Yeah, sure you can tackle this. But we don't assign tasks to new contributors anymore, since they disappear more often than not. Send us a pull request, and we'll review it.

hv7214:
@srv-twry are you still on it ? If not, then i picking this up

srv-twry:
Sure feel free. I've closed my PR

PR #500, opened 2020-03-03
fix: disallow empty plots

srv-twry:
# PR Summary
This PR disallows empty plots

- Used `[]byte{}` instead of `buff.bytes()` for NotificationPackage when Render() fails.
- [limits] Added validation of ranges by using `math.IsInf`.
- [limits/resolveLimits] Handled the case of equal `from` and `to` by adding delta to the time
- Unit test to verify that unequal `from` and `to` values are returned by `resolveLimits`

Closes/Relates #254

### DIFF
--- a/notifier/plotting.go
+++ b/notifier/plotting.go
@@ -62,7 +62,7 @@ func (notifier *StandardNotifier) buildNotificationPackagePlot(pkg NotificationP
 	}
 	if err = renderable.Render(chart.PNG, buff); err != nil {
 		buff.Reset()
-		return buff.Bytes(), err
+		return []byte{}, err
 	}
 	return buff.Bytes(), nil
 }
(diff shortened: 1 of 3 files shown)


srv-twry:
@beevee Hi,

I have done some work and would like to know if I am on the right track. I'll add the tests for different retention intervals once you confirm.

beevee:
I'm not sure whether it is the right direction, because there are no tests. To make sure that you're fixing something, please first reproduce the bug with a test.

srv-twry:
@beevee Thanks for the reply.

I have added a small unit test to test that the new changes handle the case of equal `from` and `to` values in the metrics.

However, I don't really understand how to reproduce the bug of empty plots with a test. Any pointers about where the test should go and what it should do would be appreciated.

PR #504, opened 2020-03-05
build: Upgrade to Go 1.14

beevee:
# We want fast defers, don't we?

https://golang.org/doc/go1.14

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ services:
   - redis-server
 language: go
 go:
-  - 1.13.1
+  - 1.14
 git:
   depth: 1000
 addons:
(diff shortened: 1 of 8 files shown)


beevee:
https://github.com/golang/go/commit/64cfe9fe22113cd6bc05a2c5d0cbe872b1b57860

.travis.yml
```diff
@@ -4,7 +4,7 @@ services:
   - redis-server
 language: go
 go:
-  - 1.13.1
+  - 1.14
```
litleleprikon: May we use 1.14.0 as a docker image version and travis version?

beevee: Sounds reasonable.

beevee: Doesn't work for Travis, though.

litleleprikon: It's strange as I thought that travis is using docker images with their versioning as a build environment. By the way let's merge this PR and will fix travis version later when version with patch number will appear.

Issue #501, opened 2020-03-04
Notifier killed by OOM killer on huge amount of notifications


litleleprikon:
# BUG REPORT

## What version of Moira are you using (`[binary] --version`)?

<details><pre>
$ notifier --version
moira-notifier:master.2020-02-26-11-15
</pre></details>

## What did you expect to see?
After long time that notifier was switched off and switched on again notifier should send all notifications that was created during switch off period. I would also like to have a setting that specifies how much notifications notifier can process at a time to prevent killing by OOM.

## What did you see instead?
Notifier was killed by OOM killer because it loaded all notifications into RAM.

PR #505, opened 2020-03-09
add limit to FetchNotifications function for read notifications from db

ifireice:
# PR Summary

Add params ReadBatchSize to notifier/config.go
Add 'limit' to FetchNotifications. Notification drops using "optimistic" transactions. See comment in fetchNotificationsWithLimitDo for more details.

Closes/Relates #501

### DIFF
--- a/cmd/notifier/config.go
+++ b/cmd/notifier/config.go
@@ -36,6 +36,8 @@ type notifierConfig struct {
 	Timezone string `yaml:"timezone"`
 	// Format for email sender. Default is "15:04 02.01.2006". See https://golang.org/pkg/time/#Time.Format for more details about golang time formatting.
 	DateTimeFormat string `yaml:"date_time_format"`
+	// Amount of messages notifier reads from Redis per iteration. Use notifier.NotificationsLimitUnlimited for unlimited.
+	ReadBatchSize int `yaml:"read_batch_size"`
 }
 
 type selfStateConfig struct {
@@ -80,8 +82,9 @@ func getDefault() config {
 				LastCheckDelay:          "60s",
 				NoticeInterval:          "300s",
 			},
-			FrontURI: "http://localhost",
-			Timezone: "UTC",
+			FrontURI:      "http://localhost",
+			Timezone:      "UTC",
+			ReadBatchSize: int(notifier.NotificationsLimitUnlimited),
 		},
 		Telemetry: cmd.TelemetryConfig{
 			Listen: ":8093",
@@ -119,6 +122,16 @@ func (config *notifierConfig) getSettings(logger moira.Logger) notifier.Config {
 		logger.Infof("Format '%v' parsed successfully. Current time format: %v", format, time.Now().Format(format))
 	}
 
+	readBatchSize := notifier.NotificationsLimitUnlimited
+	if config.ReadBatchSize > 0 {
+		readBatchSize = int64(config.ReadBatchSize)
+	}
+	if config.ReadBatchSize <= 0 && int64(config.ReadBatchSize) != notifier.NotificationsLimitUnlimited {
+		logger.Warningf("Current read_batch_size is %d, but it should be > 0 or %v (unlimited), value ignored",
+			config.ReadBatchSize, notifier.NotificationsLimitUnlimited)
+	}
+	logger.Infof("Current read_batch_size is %d", readBatchSize)
+
 	return notifier.Config{
 		SelfStateEnabled:  config.SelfState.Enabled,
 		SelfStateContacts: config.SelfState.Contacts,
@@ -128,6 +141,7 @@ func (config *notifierConfig) getSettings(logger moira.Logger) notifier.Config {
 		FrontURL:          config.FrontURI,
 		Location:          location,
 		DateTimeFormat:    format,
+		ReadBatchSize:     readBatchSize,
 	}
 }
 
--- a/database/redis/notification.go
+++ b/database/redis/notification.go
@@ -2,15 +2,51 @@ package redis
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
+	"github.com/moira-alert/moira/notifier"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/gomodule/redigo/redis"
 
 	"github.com/moira-alert/moira"
 	"github.com/moira-alert/moira/database/redis/reply"
 )
+
+const (
+	transactionTriesLimit = 10
+  transactionHeuristicLimit = 10000
+)
+
+// Custom error for transaction error
+type transactionError struct {}
+
+func (e transactionError) Error() string {
+	return "Transaction Error"
+}
+
+// Drops all notifications with last timestamp
+func limitNotifications(notifications []*moira.ScheduledNotification) ([]*moira.ScheduledNotification) {
+	if len(notifications) == 0 {
+		return notifications
+	}
+	i := len(notifications) - 1
+	lastTs := notifications[i].Timestamp
+
+	for ;i >= 0; i-- {
+		if notifications[i].Timestamp != lastTs {
+			break
+		}
+	}
+
+	if i == -1 {
+		return notifications
+	}
+
+	return notifications[:i+1]
+}
 
 // GetNotifications gets ScheduledNotifications in given range and full range
 func (connector *DbConnector) GetNotifications(start, end int64) ([]*moira.ScheduledNotification, int64, error) {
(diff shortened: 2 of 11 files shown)


database/redis/notification.go
```diff
@@ -98,7 +130,119 @@ func (connector *DbConnector) removeNotifications(notifications []*moira.Schedul
 }
 
 // FetchNotifications fetch notifications by given timestamp and delete it
-func (connector *DbConnector) FetchNotifications(to int64) ([]*moira.ScheduledNotification, error) {
+func (connector *DbConnector) FetchNotifications(to int64, limit int64) ([]*moira.ScheduledNotification, error) {
+	// No limit
+	if limit == 0 {
```
litleleprikon: I would like to see number as a constant `notificationsLimitUnlimited` or something like this

database/redis/notification.go
```diff
@@ -98,7 +130,119 @@ func (connector *DbConnector) removeNotifications(notifications []*moira.Schedul
...
+		return connector.fetchNotificationsNoLimit(to)
+	}
+
+	count, err := connector.notificationsCount(to)
+	if err != nil {
+		return nil, err
+	}
+
+	// Hope count will be not greater then limit when we call fetchNotificationsNoLimit
+	if count == nil {
```
litleleprikon: Do we really have situations when `count` can be nil?

database/redis/notification.go
```diff
@@ -98,7 +130,119 @@ func (connector *DbConnector) removeNotifications(notifications []*moira.Schedul
...
+		return nil, fmt.Errorf("failed to count notification in db")
+	}
+	if *count < limit/2 {
```
litleleprikon: I think that this is heuristic and this heuristic can work in wrong way. Imagine that limit is set to 2. In this case one new notification will break this heuristic.

database/redis/notification.go
litleleprikon: Even is `count` can be nil think we need to move this check to `connector.notificationsCount`

database/redis/notification.go
```diff
@@ -98,7 +130,119 @@ func (connector *DbConnector) removeNotifications(notifications []*moira.Schedul
...
+	var e *transactionError
+
+	for i := 0; i < transactionTriesLimit; i++ {
+		res, err := connector.fetchNotificationsWithLimitDo(to, limit)
+
+		if err == nil {
+			return res, nil
+		}
+
+		if errors.As(err, &e) {
```
litleleprikon: Are you trying to get pointer on pointer? probably this would be better?

```suggestion
		if errors.As(err, &transactionError{}) {
```

database/redis/notification.go
ifireice: Done

database/redis/notification.go
ifireice: For some reasons this does not work :(
```./notification.go:187:6: second argument to errors.As must be a pointer to an interface or a type implementing error```

File: database/redis/notification.go (hunk 3 above)
Comment (ifireice) [reply database/redis/notification.go]: Well, let's apply this heuristic only if limit large enough.

I know it may return a little bit more notifications than limit, but still I would like to use the heuristic because I don't want to run all the "WATCH" machinery once again.

File: database/redis/notification.go (hunk 2 above)
Comment (ifireice) [reply database/redis/notification.go]: Done

File: cmd/notifier/config.go [hunk 5]
```diff
@@ -81,6 +83,7 @@ func getDefault() config {
 			},
 			FrontURI: "http://localhost",
 			Timezone: "UTC",
+			ReadBatchSize: 0,
```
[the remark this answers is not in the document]
ifireice: Done

database/redis/notification.go
```diff
@@ -2,16 +2,48 @@ package redis
...
+	return "Transaction Error"
+}
+
+// Drops all notifications with last timestamp
+func limitNotifications(notifications []*moira.ScheduledNotification) ([]*moira.ScheduledNotification) {
+	if len(notifications) == 0 {
+		return notifications
+	}
+	i := len(notifications) - 1
+	last_ts := notifications[len(notifications) - 1].Timestamp
```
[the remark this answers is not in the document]
ifireice: Done

database/redis/notification.go
```diff
@@ -98,7 +134,123 @@ func (connector *DbConnector) removeNotifications(notifications []*moira.Schedul
...
+	var e *transactionError
+
+	for i := 0; i < transactionTriesLimit; i++ {
+		res, err := connector.fetchNotificationsWithLimitDo(to, limit)
+
+		if err == nil {
+			return res, nil
+		}
+
+		if !errors.As(err, &e) {
```
litleleprikon: I can see the root of the problem here. You are implementing an `error` interface for `*transactionError`. You need to change `Error` function receiver from pointer to concrete type and you will be able to make check like in this example:

https://play.golang.org/p/WgOafx0wdsg

database/redis/notification.go
```diff
@@ -2,16 +2,52 @@ package redis
...
 
 	"github.com/gomodule/redigo/redis"
 
 	"github.com/moira-alert/moira"
 	"github.com/moira-alert/moira/database/redis/reply"
 )
 
+const (
+	transactionTriesLimit = 10
+  transactionHeuristicLimit = 100000
```
litleleprikon: Probably this value is to high. I think that we can set it to 10k for example.

database/redis/notification.go
ifireice: Oh, I was stupid. Thank you. Now it works.

cmd/notifier/config.go
```diff
@@ -117,6 +120,14 @@ func (config *notifierConfig) getSettings(logger moira.Logger) notifier.Config {
 		logger.Infof("Format '%v' parsed successfully. Current time format: %v", format, time.Now().Format(format))
 	}
 
+	readBatchSize := notifier.NotificationsLimitUnlimited
+	if config.ReadBatchSize == 0 {
+		logger.Warning("Current read_batch_size is 0, but it should be > 0, value ignored")
+	} else {
```
litleleprikon: Do we really need `else` here?

database/redis/notification.go
```diff
@@ -98,7 +134,122 @@ func (connector *DbConnector) removeNotifications(notifications []*moira.Schedul
...
+		}
+
+		if !errors.As(err, &transactionError{}) {
+			return nil, err
+		}
+
+		time.Sleep(200 * time.Millisecond)
+	}
+
+	return nil, fmt.Errorf("Transaction tries limit exided")
```
beevee, line 193: Typo: `exceeded`?

database/redis/notification.go
```diff
@@ -98,7 +134,122 @@ func (connector *DbConnector) removeNotifications(notifications []*moira.Schedul
...
+		return nil, fmt.Errorf("failed to EXEC: %s", err)
+	}
+
+	// ZRANGEBYSCORE with LIMIT may return not all notification with last timestamp
+	// (e.g. if we have notifications with timestamps [1, 2, 3, 3, 3] and limit == 3)
+	// but ZREMRANGEBYSCORE does not have LIMIT, so will delete all notifications with last timestamp
+	// (ts 3 in our example) and then run ZRANGEBYSCORE with our new last timestamp (2 in our example).
+
+	notificationsLimited := limitNotifications(notifications)
+	last_ts := notificationsLimited[len(notificationsLimited) - 1].Timestamp
```
beevee, line 225: `lastTs` is more idiomatic.

database/redis/notification.go
```diff
@@ -98,7 +134,122 @@ func (connector *DbConnector) removeNotifications(notifications []*moira.Schedul
...
+
+	if (len(notifications) == len(notificationsLimited)) {
+		// this means that all notifications has same timestamp,
+		// we hope that all notifications with same timestamp should feat our memory
```
beevee, line 229: Typo: `fit`.

database/redis/notification.go
```diff
@@ -98,7 +134,122 @@ func (connector *DbConnector) removeNotifications(notifications []*moira.Schedul
...
+		c.Send("UNWATCH")
+		return connector.fetchNotificationsNoLimit(last_ts)
+	}
+
+	c.Send("MULTI")
+	c.Send("ZREMRANGEBYSCORE", notifierNotificationsKey, "-inf", last_ts)
+	deleted_count, err_delete := redis.Values(c.Do("EXEC"))
```
beevee, line 236: We don't do snake_case here in Go :-)

database/redis/notification.go
```diff
@@ -98,7 +134,122 @@ func (connector *DbConnector) removeNotifications(notifications []*moira.Schedul
...
+	// ZRANGEBYSCORE with LIMIT may return not all notification with last timestamp
+	// (e.g. if we have notifications with timestamps [1, 2, 3, 3, 3] and limit == 3)
+	// but ZREMRANGEBYSCORE does not have LIMIT, so will delete all notifications with last timestamp
+	// (ts 3 in our example) and then run ZRANGEBYSCORE with our new last timestamp (2 in our example).
+
+	notificationsLimited := limitNotifications(notifications)
+	last_ts := notificationsLimited[len(notificationsLimited) - 1].Timestamp
+
+	if (len(notifications) == len(notificationsLimited)) {
+		// this means that all notifications has same timestamp,
```
beevee, line 228: Typo: `have`.

database/redis/notification.go
```diff
@@ -98,7 +134,122 @@ func (connector *DbConnector) removeNotifications(notifications []*moira.Schedul
...
+
+	c.Send("MULTI")
+	c.Send("ZREMRANGEBYSCORE", notifierNotificationsKey, "-inf", last_ts)
+	deleted_count, err_delete := redis.Values(c.Do("EXEC"))
+	if err_delete != nil {
+		return nil, fmt.Errorf("failed to EXEC: %s", err_delete)
+	}
+
+	// someone has changed notifierNotificationsKey while we do our job
+	// and transaction fail (no notifications was deleted) :(
```
beevee, line 242: Typo: `were`.

cmd/notifier/config.go
```diff
@@ -119,6 +122,16 @@ func (config *notifierConfig) getSettings(logger moira.Logger) notifier.Config {
 		logger.Infof("Format '%v' parsed successfully. Current time format: %v", format, time.Now().Format(format))
 	}
 
+	readBatchSize := notifier.NotificationsLimitUnlimited
+	if config.ReadBatchSize > 0 {
+		readBatchSize = int64(config.ReadBatchSize)
+	}
+	if config.ReadBatchSize <= 0 && int64(config.ReadBatchSize) != notifier.NotificationsLimitUnlimited {
+		logger.Warningf("Current read_batch_size is %v, but it should be > 0 or %v (unlimited), value ignored",
```
litleleprikon, line 130: `%d` for integer will have higher performance and more clear.

cmd/notifier/config.go
```diff
@@ -119,6 +122,16 @@ func (config *notifierConfig) getSettings(logger moira.Logger) notifier.Config {
...
+			config.ReadBatchSize, notifier.NotificationsLimitUnlimited)
+	}
+	logger.Infof("Current read_batch_size is %v", readBatchSize)
```
litleleprikon, line 133: Same as a previous comment

database/redis/notification.go
```diff
@@ -98,7 +134,122 @@ func (connector *DbConnector) removeNotifications(notifications []*moira.Schedul
...
+}
+
+func (connector *DbConnector) notificationsCount(to int64) (int64, error) {
+	c := connector.pool.Get()
+	defer c.Close()
+
+	count, err := redis.Int64(c.Do("ZCOUNT", notifierNotificationsKey, "-inf", to))
+
+	if err != nil {
+		return 0, fmt.Errorf("failed to ZCOUNT to notificationsCount: %s", err)
```
litleleprikon, line 168: `%w` is better for errors

database/redis/notification.go
```diff
@@ -98,7 +134,122 @@ func (connector *DbConnector) removeNotifications(notifications []*moira.Schedul
...
+		// we hope that all notifications with same timestamp should fit our memory
+		c.Send("UNWATCH")
+		return connector.fetchNotificationsNoLimit(lastTs)
+	}
+
+	c.Send("MULTI")
+	c.Send("ZREMRANGEBYSCORE", notifierNotificationsKey, "-inf", lastTs)
+	deleteCount, errDelete := redis.Values(c.Do("EXEC"))
+	if errDelete != nil {
+		return nil, fmt.Errorf("failed to EXEC: %s", errDelete)
```
litleleprikon, line 238: `%w` will be better

database/redis/notification_test.go
```diff
@@ -71,8 +72,14 @@ func TestScheduledNotification(t *testing.T) {
 			So(actual, ShouldResemble, make([]*moira.ScheduledNotification, 0))
 		})
 
+		Convey("Test fetch notifications limit 0", func() {
+			actual, err := dataBase.FetchNotifications(now - 3600, 0)
+			errMsg := fmt.Errorf("limit mustn't be 0")
```
litleleprikon, line 77: I do not think that comparison with error value produced by `Errorf` is a good idea here

ifireice: Ok, I will check only that some error exists.

BTW, https://github.com/moira-alert/moira/search?q=%22So%28err%2C+ShouldResemble%2C+fmt.Errorf%22&unscoped_q=%22So%28err%2C+ShouldResemble%2C+fmt.Errorf%22 ;)

litleleprikon: There is not a good practice and I would like to reduce amount of places where this style of check is used

PR #507, opened 2020-03-10
build(mock): Update golang/mock to 1.4.1

A1bemuth:
# Update golang/mock to 1.4.1

### DIFF
--- a/mock/moira-alert/logger.go
+++ b/mock/moira-alert/logger.go
@@ -5,9 +5,8 @@
 package mock_moira_alert
 
 import (
-	reflect "reflect"
-
 	gomock "github.com/golang/mock/gomock"
+	reflect "reflect"
 )
 
 // MockLogger is a mock of Logger interface
@@ -35,6 +34,7 @@ func (m *MockLogger) EXPECT() *MockLoggerMockRecorder {
 
 // Debug mocks base method
 func (m *MockLogger) Debug(arg0 ...interface{}) {
+	m.ctrl.T.Helper()
 	varargs := []interface{}{}
 	for _, a := range arg0 {
 		varargs = append(varargs, a)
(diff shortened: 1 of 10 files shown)


mock/notifier/notifier.go
```diff
@@ -50,20 +49,6 @@ func (mr *MockNotifierMockRecorder) GetSenders() *gomock.Call {
...
-	ret0, _ := ret[0].(error)
-	return ret0
-}
-
-// InitImageStore indicates an expected call of InitImageStore
-func (mr *MockNotifierMockRecorder) InitImageStore(arg0 interface{}) *gomock.Call {
-	mr.mock.ctrl.T.Helper()
-	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "InitImageStore", reflect.TypeOf((*MockNotifier)(nil).InitImageStore), arg0)
-}
-
```
litleleprikon: Why this two methods was removed?

mock/moira-alert/logger.go
```diff
@@ -116,26 +126,30 @@ func (m *MockLogger) Fatalf(arg0 string, arg1 ...interface{}) {
 
 // Fatalf indicates an expected call of Fatalf
 func (mr *MockLoggerMockRecorder) Fatalf(arg0 interface{}, arg1 ...interface{}) *gomock.Call {
+	mr.mock.ctrl.T.Helper()
 	varargs := append([]interface{}{arg0}, arg1...)
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "Fatalf", reflect.TypeOf((*MockLogger)(nil).Fatalf), varargs...)
 }
 
-// Maintenance mocks base method
+// Info mocks base method
```
litleleprikon: Why this methods was renamed?

mock/notifier/notifier.go
A1bemuth: I've rebuilded mocks with a script, and these methods appear to be no longer needed. At least, they are not in any interface in repository.

litleleprikon: Ok, I see. I believe that someone is removed methods from interface and forgot to rebuild mocks. I just thought that you generated mocks on branch different from master.

mock/moira-alert/logger.go
A1bemuth: It seems, that the info method mock was rewritten by hand to call mock by name "Maintenance" instead of "Info". I don't see any reason for why that was done. Commit 614867417ab. 
AFAIK this has no effect on the mock's behaviour. Regenerating the mock code produced a correct version.

PR #510, opened 2020-03-12
Slave replicas read

A1bemuth:
# Implement reading patterns from slave

Implemented reading patterns from slave replicas in an attempt to balance some load. Replicas may contain stale data, but it does not seem like a problem in case of updating the pattern index. 

Did not implement things like replica staleness checking or random replicas iteration. Should consider adding those sometime.

### DIFF
--- a/database/redis/database.go
+++ b/database/redis/database.go
@@ -40,6 +40,7 @@ const (
 // DbConnector contains redis pool
 type DbConnector struct {
 	pool                 *redis.Pool
+	slavePool            *redis.Pool
 	logger               moira.Logger
 	retentionCache       *cache.Cache
 	retentionSavingCache *cache.Cache
@@ -50,7 +51,7 @@ type DbConnector struct {
 
 // NewDatabase creates Redis pool based on config
 func NewDatabase(logger moira.Logger, config Config, source DBSource) *DbConnector {
-	poolDialer := newPoolDialer(logger, config)
+	poolDialer, slaveDialer := createPoolDialers(logger, config)
 
 	pool := &redis.Pool{
 		MaxIdle:      config.ConnectionLimit,
@@ -68,9 +69,16 @@ func NewDatabase(logger moira.Logger, config Config, source DBSource) *DbConnect
 		Dial:         poolDialer.Dial,
 		TestOnBorrow: poolDialer.Test,
 	}
+	var slavePool *redis.Pool
+	if slaveDialer != nil {
+		slavePool = &*pool
+		slavePool.Dial = slaveDialer.Dial
+		slavePool.TestOnBorrow = slaveDialer.Test
+	}
 
 	return &DbConnector{
 		pool:                 pool,
+		slavePool:            slavePool,
 		logger:               logger,
 		retentionCache:       cache.New(cacheValueExpirationDuration, cacheCleanupInterval),
 		retentionSavingCache: cache.New(cache.NoExpiration, cache.DefaultExpiration),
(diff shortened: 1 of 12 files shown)


database/redis/metric.go
```diff
@@ -15,8 +15,15 @@ import (
 
 // GetPatterns gets updated patterns array
 func (connector *DbConnector) GetPatterns() ([]string, error) {
-	c := connector.pool.Get()
+	var pool *redis.Pool
```
beevee: I'm not sure this is the right place to decide. Some callers may want to use only strictly consistent master database, some may be satisfied with stale, eventually-consistent data. You propose to use slave reads for all `GetPatterns` callers, now and in the future. This can be unexpected to developers and lead to elusive bugs.

I see two possible solutions:
1. Pass a special parameter (`GetPatterns(allowStale bool)`?) to this function that allows slave reads.
2. Create a copy of this function that works with slaves (`GetPatternsAllowStale`?).
3. Create a second `DbConnector`, that uses slaves for all read (and write, if possible?) operations.

A1bemuth: Yeah, I did not really mean to leave it at that. It's a fast implementation to see if it does any difference, and if it proves useful, refactoring and rethinking would be in order.

A1bemuth:
Helmchart updates for this MR: moira-alert/helmcharts#6

database/redis/database.go
```diff
@@ -92,15 +128,23 @@ func newPoolDialer(logger moira.Logger, config Config) PoolDialer {
...
 
 	serverAddr := net.JoinHostPort(config.Host, config.Port)
 	logger.Infof("Redis: %v, DB: %v", serverAddr, config.DB)
-	return &DirectPoolDialer{
+	mainDialer = &DirectPoolDialer{
 		serverAddress: serverAddr,
 		db:            config.DB,
 		dialTimeout:   dialTimeout,
 	}
+	return
```
beevee: I feel that named returns make code less clear in this case. You return explicit values 2 times out of 3 (lines 132 and 137), and named return values just once in the end. It takes extra time to follow the code and understand where did these values come from. I suggest returning explicit values everywhere.

database/redis/database_test.go
```diff
@@ -28,3 +28,36 @@ func TestInitialization(t *testing.T) {
...
+	Convey("Allow stale", t, func() {
+		Convey("When using redis directly, returns same db", func() {
+			database := newTestDatabase(logger, emptyConfig)
+
+			staleDatabase := database.AllowStale()
+
+			So(staleDatabase, ShouldPointTo, database)
+		})
+
+		Convey("When using sentinel, returns slave db instance, retains references", func() {
```
beevee: Consider adding a test case when Sentinel is configured, but there are no slaves, only master.

beevee: I'm also curious, what happens when there is a live slave at application startup, but then it goes offline. Does this code fallback to using master without restart?

What about two slaves? If one of them goes offline, does this code reconnect to the other?

A1bemuth: Dropping a faulted connection and establishing a new one seems to be redigo's responsibility. The `Dial()` in database_dialer.go queries sentinel and returns a valid connection. My guess is that when a connection breaks, redigo will drop it from pool and call `Dial()` to obtain a new one.

database/redis/database.go
```diff
@@ -46,11 +46,13 @@ type DbConnector struct {
 	metricsCache         *cache.Cache
 	sync                 *redsync.Redsync
 	source               DBSource
+
+	slaveConnector *DbConnector
```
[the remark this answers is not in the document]
A1bemuth: Done

database/redis/database.go
A1bemuth: Done

database/redis/database_test.go
A1bemuth: Also, I've added a testcase for when the slave reads are disabled

Issue #216, opened 2019-01-22
Expression result must be state value, если введен expression вида t1 < 300 ? ERROR


borovskyav:
Это чинится если добавить ```t1 < 300 ? ERROR : OK```. Кажется нам стоит научиться более умно проверять что у нас введен правильный expression. Можно попробовать запретить вводить expression'ы без обоих операций ```?``` и ```:```.

kamaev:
> Можно попробовать запретить вводить expression'ы без обоих операций `?` и `:`
Если проверка будет строгой, то сломается паттерн триггера, который мониторит только nodata с "expression":"OK"

iurimateus:
I'll take this one

iurimateus:
I made some attempts to fix it, but I'm not sure it's they're proper.
Both work, although they don't address the evaluation of the expression, relying on parsing instead.

On `api/controller/trigger.go/saveTrigger()`
- set `validTokens` to a slice of valid/acceptable `moira.State` values and possible targets
- grab variables by constructing a govalute expression, then use the `Vars()` method
- for each variable, check if it's a `validToken`

Code: https://github.com/iurimateus/moira/commit/d2982df427ebaa15313f26648ca682369defda56
However, I don't like how it flows and the hard-coded `moira.State` values.

Another attempt:
On `expression/expression.go Evaluate()` https://github.com/moira-alert/moira/blob/683fa52a6b6a980f2b3ffa37ba6ca031186beb0a/expression/expression.go#L78-L80

- For each variable from `expr.Vars()`, take advantage of the Get method, checking for errors

https://github.com/moira-alert/moira/blob/683fa52a6b6a980f2b3ffa37ba6ca031186beb0a/expression/expression.go#L44-L45

Cleaner, and doesn't depend on copying different `moira.State` values.

Code: https://github.com/iurimateus/moira/commit/37fe4c10a9c64169c85c0ecb47c486ce39b6e21c
Problem: Not sure if its intended for the function, or if it may cause unexpected behavior.

I thought about forcing the evaluation of different parameters, so to as satisfy both sides of a ternary condition, letting the `Eval` method of `govaluate.EvaluableExpression` deal with it, but I'm not sure how to go about it.

beevee:
@iurimateus I think both your approaches fail to catch errors like `t1<300 ? ERROR`. This expression contains only valid variables, but still doesn't produce a valid state when `t1` is greater than 300.

beevee:
[...]

PR #511, opened 2020-03-15
[FEAT] Check expression validity on trigger update and save

hv7214:
# PR Summary

- Added `checkExpressionValidity` function which checks for the validity of the expression.
- This function user `expression.go/Evaluate` function.
- Added tests for same - *wrong syntax* & *wrong state*. 

**Related issue**
#216

### DIFF
--- a/api/controller/trigger.go
+++ b/api/controller/trigger.go
@@ -8,6 +8,7 @@ import (
 	"github.com/moira-alert/moira/api"
 	"github.com/moira-alert/moira/api/dto"
 	"github.com/moira-alert/moira/database"
+	"github.com/moira-alert/moira/expression"
 )
 
 // UpdateTrigger update trigger data and trigger metrics in last state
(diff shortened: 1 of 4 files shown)


api/controller/trigger.go
```diff
@@ -27,6 +29,9 @@ func saveTrigger(dataBase moira.Database, trigger *moira.Trigger, triggerID stri
 	if err := dataBase.AcquireTriggerCheckLock(triggerID, 10); err != nil {
 		return nil, api.ErrorInternalServer(err)
 	}
+	if err := checkExpressionValidity(trigger); err != nil {
```
beevee: Same check is already done in https://github.com/moira-alert/moira/blob/a19b756b9ee9ef4f5ac3547e88e1dae688a2bb81/api/dto/triggers.go#L142

Why do you think it's going to help to make the same check twice?

hv7214:
Closing this pull request. Will make new PR.

PR #516, opened 2020-03-17
[BREAKING CHANGE] Removed paging mechanism

hv7214:
# PR Summary

**Details**

1. Now `GetTriggerEvents` returns all events in one go.
2. To do this, first calculated total event count and then passed it `GetNotificationEvents` to get all events.
3. Updated tests accordingly.
4. Db mock is also updated accordingly.

**Additional information**

This PR results in frontend breakage.
Frontend PR : https://github.com/moira-alert/web2.0/pull/305 
 
**Related issue**
#363

### DIFF
--- a/api/controller/events.go
+++ b/api/controller/events.go
@@ -7,16 +7,14 @@ import (
 )
 
 // GetTriggerEvents gets trigger event from current page and all trigger event count
-func GetTriggerEvents(database moira.Database, triggerID string, page int64, size int64) (*dto.EventsList, *api.ErrorResponse) {
-	events, err := database.GetNotificationEvents(triggerID, page*size, size-1)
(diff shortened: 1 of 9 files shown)


hv7214:
@litleleprikon please have a look here: https://github.com/moira-alert/web2.0/pull/305#issuecomment-599933920

beevee:
Yes, we should definitely preserve the backend paging mechanism. There is no need to drop this functionality, considering that there may be other API clients that rely on it, not just web UI. In https://github.com/moira-alert/web2.0/pull/305#issuecomment-599933920 I meant getting rid of paging on frontend side, not on backend side.

I'm closing this.

hv7214:
@beevee I see will figure out another way on the frontend side

PR #518, opened 2020-03-21
[FIX] Expression parsing

hv7214:
# PR Summary

1. Fixes expression parsing issues outlined below.
2. Associated tests are added.

Issues :
1. t1 > 10 ? ERROR
2. t1 > 10 ? ERROR : ()
3. t1 > 10 ? ERROR : foo
4. t1 > 10 ? foo : ERROR

**Related issue**

#216

### DIFF
--- a/expression/expression.go
+++ b/expression/expression.go
@@ -144,14 +144,15 @@ func getUserExpression(triggerExpression *TriggerExpression) (*govaluate.Evaluab
 		}
 		return nil, err
 	}
-	if err := triggerExpression.expressionExtraCheckLayer(expr.Vars()); err != nil {
+	if err := triggerExpression.validateUserExpression(expr.Vars()); err != nil {
 		return nil, err
 	}
 	exprCache.Add(*triggerExpression.Expression, expr, cache.NoExpiration)
 	return expr, nil
 }
 
-func (triggerExpression *TriggerExpression) expressionExtraCheckLayer(vars []string) error {
+// This expression validation catches those errors which are ignored by govaluate
+func (triggerExpression *TriggerExpression) validateUserExpression(vars []string) error {
 	for _, v := range vars {
 		if _, err := triggerExpression.Get(v); err != nil {
 			return err
--- a/expression/expression.go
+++ b/expression/expression.go
@@ -98,7 +98,7 @@ func getExpression(triggerExpression *TriggerExpression) (*govaluate.EvaluableEx
 		if triggerExpression.Expression == nil || *triggerExpression.Expression == "" {
 			return nil, fmt.Errorf("trigger_type set to expression, but no expression provided")
 		}
-		return getUserExpression(*triggerExpression.Expression)
+		return getUserExpression(triggerExpression)
 	}
 	return getSimpleExpression(triggerExpression)
 }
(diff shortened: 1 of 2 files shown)


hv7214:
**`Please ignore above mention`**

expression/expression.go
```diff
@@ -131,19 +131,65 @@ func getSimpleExpression(triggerExpression *TriggerExpression) (*govaluate.Evalu
...
 	return expr, nil
 }
+
+func (triggerExpression *TriggerExpression) expressionExtraCheckLayer(vars []string) error {
+	for _, v := range vars {
+		if _, err := triggerExpression.Get(v); err != nil {
+			return err
+		}
+	}
+	if isTernaryExpression(*triggerExpression.Expression) && !checkforColon(*triggerExpression.Expression) || !checkForEmptyStates(*triggerExpression.Expression) {
```
litleleprikon: I do not think that this check is complex enough to check all kinds of ternary expressions. Let's start with easy example: `?:?`. Will it pass your check? Yes. Is it correct expression? No. Check the example https://play.golang.org/p/13uxKuoi-cF

Think about nested ternary expressions especially invalid ones. Your solution will not handle it correctly. 

I propose to not to check expressions by ourselves and pass it to specialised library. I can see that call of `govaluate.NewEvaluableExpression` will check expression for correctness so probably calling it will be better that checking by ourselves.

expression/expression.go
```diff
@@ -98,7 +98,7 @@ func getExpression(triggerExpression *TriggerExpression) (*govaluate.EvaluableEx
 		if triggerExpression.Expression == nil || *triggerExpression.Expression == "" {
 			return nil, fmt.Errorf("trigger_type set to expression, but no expression provided")
 		}
-		return getUserExpression(*triggerExpression.Expression)
+		return getUserExpression(triggerExpression)
```
litleleprikon: One point about check that I cannot stop thinking about is a performance. I want to see this functionality covered with benchmarks with various expressions tested(simple, complex, nested, nested with many layers) and comparison of performance before check implemented and after

expression/expression.go
hv7214: @litleleprikon that's why i've put the function after this call `govaluate.NewEvaluableExpression()`. So basically this function acts like another filter above the govaluate validation, which detects ignored errors of evaluate.

expression/expression.go
```diff
@@ -131,19 +131,65 @@ func getSimpleExpression(triggerExpression *TriggerExpression) (*govaluate.Evalu
...
-
-	exprCache.Add(triggerExpression, expr, cache.NoExpiration)
+	if err := triggerExpression.expressionExtraCheckLayer(expr.Vars()); err != nil {
+		return nil, err
+	}
+	exprCache.Add(*triggerExpression.Expression, expr, cache.NoExpiration)
 	return expr, nil
 }
+
+func (triggerExpression *TriggerExpression) expressionExtraCheckLayer(vars []string) error {
```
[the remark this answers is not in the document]
hv7214: sure

expression/expression.go
hv7214: Yeah, I agree with you, these needs benchmarks log.

hv7214: I will add those after the logic consolidation

kissken:
work in https://github.com/moira-alert/moira/pull/957

PR #520, opened 2020-03-26
fix(api): enable reporting of why notifications were turned off

### LINKED ISSUE #469: Report who and why turned off notifications

idoqo:
# PR Summary
When Moira notifications are turned off, we show a default message (`dto.ErrorMessage`) and this isn't always helpful. Instead: 
- Let users describe their purpose when they disable notifications from the web UI. This is possible as the `NotifierState` struct already supports both `State` and `Message`.
- Log a more helpful message when Moira turns itself off but is still able to access the database (e.g `filter` receives no metrics for longer than last_metric_received_delay)

# Additional information
We need to add a text field to the `notifications` page for users to describe their purpose when disabling notifications. I plan to do that once the API changes are okay and merged.

Closes #469

### DIFF
--- a/database/redis/selfstate.go
+++ b/database/redis/selfstate.go
@@ -47,28 +47,43 @@ func (connector *DbConnector) GetRemoteChecksUpdatesCount() (int64, error) {
 }
 
 // GetNotifierState return current notifier state: <OK|ERROR>
-func (connector *DbConnector) GetNotifierState() (string, error) {
+func (connector *DbConnector) GetNotifierState() (string, string, error) {
 	c := connector.pool.Get()
 	defer c.Close()
-	ts, err := redis.String(c.Do("GET", selfStateNotifierHealth))
-	if err == redis.ErrNil {
-		ts = moira.SelfStateOK
-		err = connector.SetNotifierState(ts)
-	} else if err != nil {
-		ts = moira.SelfStateERROR
+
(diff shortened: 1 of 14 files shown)


database/redis/selfstate.go
```diff
@@ -47,28 +47,43 @@ func (connector *DbConnector) GetRemoteChecksUpdatesCount() (int64, error) {
...
-	ts, err := redis.String(c.Do("GET", selfStateNotifierHealth))
-	if err == redis.ErrNil {
-		ts = moira.SelfStateOK
-		err = connector.SetNotifierState(ts)
-	} else if err != nil {
-		ts = moira.SelfStateERROR
+
+	var state, message string
+	stateArgs := []interface{} {selfStateNotifierHealth, selfStateNotifierMessage }
+	values, err := redis.Strings(c.Do("MGET", stateArgs...))
```
idoqo: I've used `MGET` (and `MSET` for SetNotifierState too) as it feels better than getting State and Message separately and using multiple if-statements to check for their errors. I'm not so confident about it though, and would like to know what you think.

PR #527, opened 2020-04-06
feat(notifier): Add metrics for every possible state transition

beevee:
# PR Summary

Additional metrics to help analyzing spikes of delivered notifications.

Metric names: `...events.bystate.OK_to_NODATA`, etc.

Frankly, I don't understand why we have meters instead of counters here. I added these metrics as meters to maintain uniformity.

### DIFF
--- a/notifier/events/event.go
+++ b/notifier/events/event.go
@@ -43,7 +43,15 @@ func (worker *FetchEventsWorker) Start() {
 						}
 						continue
 					}
+
 					worker.Metrics.EventsReceived.Mark(1)
+					stateMeterName := event.OldState.String() + "_to_" + event.State.String()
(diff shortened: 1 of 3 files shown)


notifier/events/event.go
```diff
@@ -43,7 +43,15 @@ func (worker *FetchEventsWorker) Start() {
 						}
 						continue
 					}
+
 					worker.Metrics.EventsReceived.Mark(1)
+					stateMeterName := event.OldState.String() + "_to_" + event.State.String()
```
litleleprikon, line 48: What do you think about states as a nodes of metric? Like `OK.to.ERROR` instead of `OK_to_ERROR`.

beevee: I thought of this, but it sounds inconvenient. I can't imagine a scenario in which it would be necessary.

PR #529, opened 2020-04-06
feat: test on a non-primary database

Nixolay:


database/redis/database.go
```diff
@@ -108,6 +108,12 @@ func NewDatabase(logger moira.Logger, config Config, source DBSource) *DbConnect
 	return connector
 }
 
+// NewTestDatabase use it only for tests
+func NewTestDatabase(logger moira.Logger, config Config, source DBSource) *DbConnector {
```
beevee, line 112: I think you should hardcode `source` here, and not pass it as a parameter. It should always be the same for tests.

database/redis/database.go
```diff
@@ -108,6 +108,12 @@ func NewDatabase(logger moira.Logger, config Config, source DBSource) *DbConnect
...
+	config.DB = 1
+	return NewDatabase(logger,config,source)
```
beevee, line 114: Please, add spaces between arguments.

database/redis/database.go
Nixolay: fixed

database/redis/database.go
Nixolay: fixed

database/redis/database_test.go
```diff
@@ -5,24 +5,13 @@ import (
...
 
-var config = Config{Port: "6379", Host: "0.0.0.0"}
-var emptyConfig = Config{}
-var testSource = DBSource("test")
-
-// use it only for tests
-func newTestDatabase(logger moira.Logger, config Config) *DbConnector {
-	return NewDatabase(logger, config, testSource)
-}
-
```
litleleprikon, line 20: I am not sure that we should remove this configs from test files and place in database.go. I think that we need just to place this configs in both modules that use it. Of course this is copy-paste and do not follow that DRY principle but I think that it's ok for tests. Placing this configs in database.go can lead to misunderstand the purpose of this configs by contributors and using it in another places of codebase.

Nixolay: fixed

database/redis/database.go
```diff
@@ -37,6 +37,14 @@ const (
...
 )
 
+// Test data for configuration
+const (
+	db                  = 1
+	testSource DBSource = "test"
+)
+
+var testConfig = Config{Port: "6379", Host: "0.0.0.0", DB: db}
+
```
litleleprikon, line 47: My previous comment was that we should not have this config in this file. This config should appear only in test files

Nixolay: fixed

PR #531, opened 2020-04-13
Mark all dangerous Graphite functions as such

### LINKED ISSUE #182: Mark all dangerous Graphite functions as such (show a warning)
Если в поле есть функция summarize (как пример, нужно составить список всех неугодных функций), подсвечивать желтым, показывать тултип с описанием, почему так. Надо составить список таких функций. Допилить документацию.

ToDo:
- [ ] Backend
- [ ] Frontend
- [ ] Doc

Nixolay:
Close #182

### DIFF
--- a/api/dto/target.go
+++ b/api/dto/target.go
@@ -1,0 +1,251 @@
+package dto
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/go-graphite/carbonapi/expr/functions"
+	"github.com/go-graphite/carbonapi/expr/metadata"
+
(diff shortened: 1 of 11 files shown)


api/dto/target.go
```diff
@@ -0,0 +1,187 @@
+package dto
```
litleleprikon, line 1: I do not think that `dto` package is a good place for this logic. In `dto` should be placed only request-response data schemes and only validation of this schemas.

litleleprikon (changes requested):

litleleprikon (approved):

PR #554, opened 2020-06-22
feat(checker): Allow targets be single if it not declared

litleleprikon:
Targets in trigger can be marked as a "Single" what is mean that in
this target will be only one metric. If target have one metric and it is
not marked as "Single" check of this trigger will set trigger to
exception state as it was expected to have more than one metric in this
target.

This commit changes behavior of checker to allow targets not marked as
alone to have only one metric.

Relates #428

### DIFF
--- a/checker/metrics/conversion/trigger_metrics.go
+++ b/checker/metrics/conversion/trigger_metrics.go
@@ -62,19 +62,18 @@ func NewTriggerMetricsWithCapacity(capacity int) TriggerMetrics {
 // Populate is a function that takes TriggerMetrics and populate targets
 // that is missing metrics that appear in another targets except the targets that have
 // only alone metrics.
-func (m TriggerMetrics) Populate(lastCheck moira.CheckData, from int64, to int64) TriggerMetrics {
+func (m TriggerMetrics) Populate(lastCheck moira.CheckData, declaredAloneMetrics map[string]bool, from int64, to int64) TriggerMetrics {
+	// This one have all metrics that should be in final TriggerMetrics.
+	// This structure filled with metrics from last check,
+	// current received metrics alone metrics from last check.
 	allMetrics := make(map[string]map[string]bool, len(m))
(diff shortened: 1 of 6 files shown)


checker/metrics/conversion/trigger_metrics_test.go
```diff
@@ -276,11 +281,46 @@ func TestTriggerMetrics_Populate(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "in last check exist metric with name of alone metric",
+			m: TriggerMetrics{
```
Nixolay, line 286: It seems to me that even here it is worth using the full name of the variable. Maybe it is worth changing the structure?

checker/metrics/conversion/trigger_metrics.go
```diff
@@ -132,7 +133,7 @@ func (m TriggerMetrics) FilterAloneMetrics() (TriggerMetrics, map[string]metricS
 
 // Diff is a function that returns a map of target names with metric names that are absent in
 // current target but appear in another targets.
-func (m TriggerMetrics) Diff() map[string]map[string]bool {
+func (m TriggerMetrics) Diff(declaredAloneMetrics map[string]bool) map[string]map[string]bool {
```
Nixolay, line 136: I think the fully qualified identifier names will make it much easier for developers to understand what the system does and how to fix or extend the source code for new purposes. I'm not sure if in this case it is worth neglecting

checker/metrics/conversion/trigger_metrics_test.go
litleleprikon: It's an auto-generated test. The name of structure attribute here taken from the name of tested function receiver.

checker/metrics/conversion/trigger_metrics.go
litleleprikon: Which identifier is not clear in this function?

Nixolay: `m TriggerMetrics`

borovskyav (approved):

PR #556, opened 2020-06-26
Return false as default pass value in handlePrepareError

litleleprikon:
In checker handlePrepareError returns a boolean called pass which notify
caller should it continue or should it return with error immediately.
By default true was returned. Default changed to false.

Relates #428


litleleprikon:
I am not sure about the reasons, but this commit is already in master. I am absolutely have no idea how this PR is still open. I will close it. If anyone have any reasons to reopen it then you are welcome.

Commit in master: https://github.com/moira-alert/moira/commit/76053ee33074462a64d86da0a28e515c5cded565

PR #580, opened 2020-10-09
fix(notifier): Fixed sending of plots in notifications

zhelyabuzhsky:
# Fixed sending of plots in notifications

Plots were not sent to Slack and Telegram. The problem was iterating over the keys in the map `metricsData`.

### DIFF
--- a/notifier/plotting.go
+++ b/notifier/plotting.go
@@ -32,9 +32,25 @@ func (err errFetchAvailableSeriesFailed) Error() string {
 	return fmt.Sprintf("Failed to fetch both realtime and stored data: [realtime]: %s, [stored]: %s", err.realtimeErr, err.storedErr)
 }
 
-// buildNotificationPackagePlot returns bytes slice containing package plot
-func (notifier *StandardNotifier) buildNotificationPackagePlots(pkg NotificationPackage) ([][]byte, error) {
+// buildTriggerPlots returns bytes slices containing trigger plots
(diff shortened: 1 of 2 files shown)


notifier/plotting.go
```diff
@@ -32,9 +32,26 @@ func (err errFetchAvailableSeriesFailed) Error() string {
 	return fmt.Sprintf("Failed to fetch both realtime and stored data: [realtime]: %s, [stored]: %s", err.realtimeErr, err.storedErr)
 }
 
-// buildNotificationPackagePlot returns bytes slice containing package plot
-func (notifier *StandardNotifier) buildNotificationPackagePlots(pkg NotificationPackage) ([][]byte, error) {
+// buildTriggerPlots returns bytes slices containing trigger plots
+func buildTriggerPlots(trigger *moira.Trigger, metricsData map[string][]metricSource.MetricData, plotTemplate *plotting.Plot) ([][]byte, error) {
 	result := make([][]byte, 0)
+	for targetName := range metricsData {
+		metrics := metricsData[targetName]
```
litleleprikon, line 39: It's better to use `for targetName, metrics := range metricsData` here

litleleprikon (approved):

notifier/plotting.go
litleleprikon, line 39: It's better to use for targetName, metrics := range metricsData here

litleleprikon (changes requested):

notifier/plotting.go
zhelyabuzhsky: done

litleleprikon (approved):

PR #585, opened 2020-10-20
Add support of Slack user-group mentioning in the alert message

ArXa1L:
Hello, 
My team has a single channel for alerts across all our applications. For each application we added a separate Slack user-group (`@gitlab_duty`/`@teamcity_duty` for example) that contains one or several engineers on duty and we would like to mention these groups in the alert message. 

Moira uses Slack API `chat.postMessage` for sending alert messages and that requires us to specify that user group by using the rather complicated syntax `<!Subteam^SAZ94GDB8>` (see https://api.slack.com/reference/surfaces/formatting#mentioning-groups). We can't get user-group ID from the Slack Desktop App and we have to use Slack API in order to do that.

There is another way: we can specify additional property `Parse` with value `full` in PostMessageParameters which change how Slack treats the message (see https://api.slack.com/methods/chat.postMessage#formatting). In my opinion this way is more user-friendly and the required changes are minor.

### DIFF
--- a/senders/slack/slack.go
+++ b/senders/slack/slack.go
@@ -180,6 +180,8 @@ func (sender *Sender) sendMessage(message string, contact string, triggerID stri
 		AsUser:    useDirectMessaging,
 		IconEmoji: emoji,
 		Markdown:  true,
+		Parse: "full",
+		LinkNames: 1,
 	}
 	sender.logger.Debugf("Calling slack with message body %s", message)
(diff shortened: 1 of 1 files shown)


ArXa1L:
Unofficially, the 'Parse'-argument is deprecated and will be removed soon. I added an additional argument that is responsible for handling links.

ArXa1L:
According to unofficial information, the "Parse" can be deprecated and will be removed soon. Just in case, I added one more property that changes in similar way how Slack API processes the message (see https://api.slack.com/methods/chat.postMessage#arg_link_names)

beevee (approved):

beevee (approved):

PR #601, opened 2021-02-10
bugfix: get trigger/check method has params via url, not body

androndo:
# PR Summary

An error appears when saving some edited trigger in the web interface.

This method [call](https://github.com/moira-alert/web2.0/commit/e6ff286b2e53f41cafa5111e5448e12fae414625#diff-856b28a6b72ebb23dad388c2cc5cc00c4e3014d7b3d97c33b514201d61bf0433R360) in web2.0.

### DIFF
--- a/api/handler/triggers.go
+++ b/api/handler/triggers.go
@@ -1,6 +1,7 @@
 package handler
 
 import (
+	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


androndo:
Smartly, fix this at web and add api docs on method

Issue #614, opened 2021-02-26
Autodisabling blocked subscriptions


androndo:
Today some subscriptions may repeat failed sendings about 50+ times, because:
* telegram bot blocked
* slack channel archieved
* wrong contact address
* etc.

But some reasons is ok: timeouts or others (?), we need pass them.

This sending attempts generate unneccessary activity and trash collected at logs.

Possible way:
Add some euristic to check notification fail reasons and attempts count, and disable subscription automatically.
Add to admin page: enable/disable this feature and abillity to restore (enable all disabled) in case of fuckups.
Send banned count metric.

PR #615, opened 2021-02-26
Feature/detect broken contacts

androndo:
# PR Summary

Additional information

Relates #614

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -6,6 +6,12 @@ type SenderBrokenContactError struct {
 	SenderError error
 }
 
-func (e *SenderBrokenContactError) Error() string {
+func NewSenderBrokenContactError(senderError error) SenderBrokenContactError {
+	return SenderBrokenContactError{
+		SenderError: senderError,
+	}
+}
+
+func (e SenderBrokenContactError) Error() string {
 	return e.SenderError.Error()
 }
--- a/errors.go
+++ b/errors.go
@@ -1,4 +1,4 @@
-package notifier
+package moira
 
 //SenderBrokenContactError means than sender has no way to send message to contact.
 // Maybe receive contact was deleted, blocked or archived.
--- a/errors.go
+++ b/errors.go
@@ -7,5 +7,5 @@ type SenderBrokenContactError struct {
 }
 
 func (e *SenderBrokenContactError) Error() string {
-	return e.Error()
+	return e.SenderError.Error()
 }
--- a/notifier/notifier_test.go
+++ b/notifier/notifier_test.go
@@ -117,6 +117,27 @@ func TestFailSendEvent(t *testing.T) {
 	time.Sleep(time.Second * 2)
 }
 
+func TestNoResendForSendToBrokenContact(t *testing.T) {
+	configureNotifier(t)
+	defer afterTest()
+
+	var eventsData moira.NotificationEvents = []moira.NotificationEvent{event}
+
+	pkg := NotificationPackage{
+		Events: eventsData,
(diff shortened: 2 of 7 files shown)


errors.go
```diff
@@ -0,0 +1,11 @@
+package moira
+
+//SenderBrokenContactError means than sender has no way to send message to contact.
+// Maybe receive contact was deleted, blocked or archived.
+type SenderBrokenContactError struct {
+	SenderError error
+}
+
+func (e *SenderBrokenContactError) Error() string {
```
litleleprikon, line 9: Let's use concrete receiver instead of pointer

errors.go
```diff
@@ -0,0 +1,11 @@
+package moira
+
+//SenderBrokenContactError means than sender has no way to send message to contact.
+// Maybe receive contact was deleted, blocked or archived.
+type SenderBrokenContactError struct {
+	SenderError error
+}
```
litleleprikon, line 7: I wish to see a `NewSenderBrokenContactError` function here that will be used in senders to create a new errors instead of creating new errors via structure literal

litleleprikon (changes requested):

errors.go
androndo: fixed

errors.go
androndo: fixed

notifier/notifier_test.go
```diff
@@ -117,6 +117,27 @@ func TestFailSendEvent(t *testing.T) {
 	time.Sleep(time.Second * 2)
 }
 
+func TestNoResendForSendToBrokenContact(t *testing.T) {
```
litleleprikon, line 120: Let's assume this test as a slow test and will skip it on short testing
```go
    if testing.Short() {
        t.Skip("skipping test in short mode.")
    }
```

senders/telegram/send.go
```diff
@@ -37,9 +37,15 @@ func (sender *Sender) SendEvents(events moira.NotificationEvents, contact moira.
 	sender.logger.Debugf("Calling telegram api with chat_id %s and message body %s", contact.Value, message)
 	chat, err := sender.getChat(contact.Value)
 	if err != nil {
+		if ok, errorBrokenContact := checkBrokenContactError(sender.logger, err); ok {
```
litleleprikon, line 40: It's pretty strange to see that we raise an error if `ok` is true. Probably we need to invert bool value in return of function and invert condition here?

litleleprikon: Additionally: do we really need this bool value? probably we can just return `SenderBrokenContactError` if contact is broken and nil otherwise. And this condition will become just 

```go
if errorBrokenContact := checkBrokenContactError(sender.logger, err); errorBrokenContact != nil {
```

litleleprikon: Or even better solution. Let's the function `checkBrokenContactError` will return `SenderBrokenContactError` if contact is broken or original error otherwise. And here we can remove this second if and use just

```go
if err != nil {
	return checkBrokenContactError(sender.logger, err)
}
```

litleleprikon (changes requested):

senders/telegram/send.go
androndo: looks good)

notifier/notifier_test.go
androndo: fixed

senders/telegram/send.go
androndo: fixed

litleleprikon (approved):

PR #616, opened 2021-03-04
Feature/db layer metrics

androndo:
# PR Summary

Marked as draft until wait this [PR](https://github.com/hexdigest/gowrap/pull/33) to go-wrap generator.

Closes/Relates #issue

### DIFF
--- a/cmd/notifier/main.go
+++ b/cmd/notifier/main.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/moira-alert/moira"
 	"github.com/moira-alert/moira/cmd"
+	db "github.com/moira-alert/moira/database"
 	"github.com/moira-alert/moira/database/redis"
 	logging "github.com/moira-alert/moira/logging/zerolog_adapter"
 	"github.com/moira-alert/moira/metrics"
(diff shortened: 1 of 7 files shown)


cmd/notifier/main.go
```diff
@@ -75,7 +76,8 @@ func main() {
 
 	notifierMetrics := metrics.ConfigureNotifierMetrics(telemetry.Metrics, serviceName)
 	databaseSettings := config.Redis.GetSettings()
-	database := redis.NewDatabase(logger, databaseSettings, redis.Notifier)
+	redis := redis.NewDatabase(logger, databaseSettings, redis.Notifier)
+	database := db.NewDatabaseWithMetrics(redis, "redis", &telemetry.Metrics)
```
litleleprikon, line 80: `"redis"` is not a good prefix for metrics. We have a metrics prefix that is set in config files. Each service adds its own name to prefix. Let's use this as prefix

metrics/gen_templates/moira-metrics
```diff
@@ -0,0 +1,45 @@
+import (
```
litleleprikon, line 1: Probably we need to add an extension to file

litleleprikon (changes requested):

PR #627, opened 2021-04-22
fix http schema escaping when build url in webhook sender

androndo:
# PR Summary

Bug with webhook: url has encoded slashes if contact value looks like `http://ya.ru/`.

Closes/Relates #issue


litleleprikon (changes requested):

androndo:
tested this fix on staging

litleleprikon (approved):

PR #640, opened 2021-06-15
Implement redis clustering mode

litleleprikon:
# PR Summary

From this point we start migration from redigo redis driver to go-redis
that will allow us to use redis clustering and mostly maintained than
predecessor.


zhelyabuzhsky:
It will be reworked

PR #646, opened 2021-08-05
feat: Switch off exceptions on database freezes

litleleprikon:
# PR Summary

Ignore exceptions that generated by redis timeouts. This exceptions do
not have any value for users but can be generated a lot and harm users.


zhelyabuzhsky:
It will be reworked here https://github.com/moira-alert/moira/pull/647

PR #648, opened 2021-08-27
No more notifications about exceptions

zhelyabuzhsky:
# No more notifications about exceptions

There are no more notifications on transition ANY STATE -> EXCEPTION and EXCEPTION -> ANY STATE. These notifications do not have any value for users.

## Notes:
Maybe it would be better to send notifications EXCEPTION -> NOT OK?


Dexes (approved):

zhelyabuzhsky:
Not relevant anymore

PR #687, opened 2021-11-17
Fixed contacts on redis cluster

zhelyabuzhsky:
There were two problems in `GetContacts`:
1. used `Keys` command that [not recommended](https://redis.io/commands/KEYS) to use in production
2. keys searched for only on one node (which, in the case of a cluster, led to incorrect results)

Also, improved the `flush` method to work on the Redis cluster correctly.

### DIFF
--- a/database/redis/contact.go
+++ b/database/redis/contact.go
@@ -1,6 +1,7 @@
 package redis
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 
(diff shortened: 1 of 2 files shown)


dmitryanchikov (approved):
Nothing jumps out to me. Do you know any other places where we need to revisit methods to place those switches with `ClusterClient` override?

zhelyabuzhsky:
> Do you know any other places where we need to revisit methods to place those switches with `ClusterClient` override?

Yes ) Stay tuned 😀

zhelyabuzhsky:
https://github.com/moira-alert/moira/pull/689

KiskachiMaria (approved):

PR #688, opened 2021-11-18
Removed unnecessary newTestDatabase function

zhelyabuzhsky:
There were two places where you needed to change the connection string to the test database. It was not very elegant.

### DIFF
--- a/database/redis/database.go
+++ b/database/redis/database.go
@@ -135,7 +135,19 @@ func (connector *DbConnector) manageSubscriptions(tomb *tomb.Tomb, channel strin
 
 // Deletes all the keys of the DB, use it only for tests
 func (connector *DbConnector) flush() {
-	(*connector.client).FlushDB(connector.context)
+	client := *connector.client
+
+	switch c := client.(type) {
+	case *redis.ClusterClient:
+		err := c.ForEachMaster(connector.context, func(ctx context.Context, shard *redis.Client) error {
+			return shard.FlushDB(ctx).Err()
+		})
+		if err != nil {
+			return
+		}
+	default:
+		(*connector.client).FlushDB(connector.context)
+	}
 }
 
 // Get key ttl, use it only for tests
--- a/database/redis/database.go
+++ b/database/redis/database.go
@@ -79,7 +79,14 @@ func NewDatabase(logger moira.Logger, config Config, source DBSource) *DbConnect
 
 // NewTestDatabase use it only for tests
 func NewTestDatabase(logger moira.Logger) *DbConnector {
-	return NewDatabase(logger, Config{Addrs: []string{"0.0.0.0:6379"}}, testSource)
+	return NewDatabase(logger, Config{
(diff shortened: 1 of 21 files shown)


database/redis/selfstate_test.go
```diff
@@ -65,7 +66,7 @@ func TestSelfCheckWithWritesNotInChecker(t *testing.T) {
 
 func testSelfCheckWithWritesInDBSource(t *testing.T, dbSource DBSource) {
 	logger, _ := logging.GetLogger("dataBase")
-	dataBase := NewDatabase(logger, config, dbSource)
+	dataBase := NewTestDatabase(logger)
```
dmitryanchikov, line 69: Probably you'd want to add like you add above for `Checker`:
```
	dataBase.source = dbSource 
```

database/redis/database.go
```diff
@@ -79,7 +79,14 @@ func NewDatabase(logger moira.Logger, config Config, source DBSource) *DbConnect
 
 // NewTestDatabase use it only for tests
 func NewTestDatabase(logger moira.Logger) *DbConnector {
-	return NewDatabase(logger, Config{Addrs: []string{"0.0.0.0:6379"}}, testSource)
+	return NewDatabase(logger, Config{
+		Addrs: []string{"0.0.0.0:6379"},
+	}, testSource)
+}
```
dmitryanchikov, line 85: I see that some tests rely on existing dbSource in test daatbase object, would it make more sense to add it as argument to `NewTestDatabase` func with `default=testSource`?

database/redis/database_test.go
```diff
@@ -4,25 +4,16 @@ import (
...
-
-func newTestDatabase(logger moira.Logger, config Config) *DbConnector {
-	return NewDatabase(logger, config, testSource)
-}
-
 func TestNewDatabase(t *testing.T) {
 	Convey("NewDatabase should return correct DBConnector", t, func() {
 		logger, _ := logging.ConfigureLog("stdout", "info", "test", true) // nolint: govet
-		database := NewDatabase(logger, config, testSource)
+		database := NewTestDatabase(logger)
```
dmitryanchikov, line 16: `dbSource=testSource` as well?

dmitryanchikov (commented):
Looks great, just a few questions

database/redis/database_test.go
dmitryanchikov: Disregard, I see tht internally `NewTestDatabase` already set `testSource`, so it is good

database/redis/database.go
zhelyabuzhsky: There are not default arguments in Go )

database/redis/selfstate_test.go
zhelyabuzhsky: fixed

dmitryanchikov (approved):

KiskachiMaria (approved):

PR #689, opened 2021-11-18
Fixed subscription_test.go on Redis Cluster

zhelyabuzhsky:


dmitryanchikov (commented):
Looks like it is becoming common pattern. Should we extract universal client scan to shared function (to use it everywhere else)? Something like:
```
universalScanByPattern(ctx context.Context, client redis.UniversalClient, pattern string)
```
Not necessarily now, but at some point if you think it tends to be common as well

zhelyabuzhsky:
> Should we extract universal client scan to shared function (to use it everywhere else)?

Maybe you are right, but as you see, there are only two places now (1 of them is in the tests).

It seems to me that the time has not come yet ;)

dmitryanchikov (approved):
Sounds good to me

dmitryanchikov (approved):

dmitryanchikov (approved):

PR #683, opened 2021-11-08
Fixed trigger list manipulations on redis cluster

zhelyabuzhsky:
There was a message like this in console:

```
checker ERR NODATA check failed: failed to get triggers-list: CROSSSLOT Keys in request don't hash to the same slot module=checker
```

Error was in `GetLocalTriggerIDs` (`SDiff`)

PR #690, opened 2021-11-26
Added command to transform keys to work on Redis Cluster

zhelyabuzhsky:
There were 5 renames of keys (2 [here](https://github.com/moira-alert/moira/pull/679),  1 [here](https://github.com/moira-alert/moira/pull/684), 2 [here](https://github.com/moira-alert/moira/pull/683)) to migrate to Redis Cluster

In this PR added a command that renames existing keys in the database.

### How to use it

1. compile cli
```bash
$ go build -o build/cli github.com/moira-alert/moira/cmd/cli
```

2. run command forwards
```bash
$ ./build/cli -config local/cli.yml -move-to-cluster-forwards
```

```bash
# time ./cli -config cli.yml -move-to-cluster-forwards
2021-12-01 09:14:23.604 cli INF Moving to cluster forwards module=cli
2021-12-01 09:14:23.606 cli INF renameAnyTagsSubscriptionsKeyForwards done module=cli
2021-12-01 09:14:23.606 cli INF renameTriggersListKeyForwards done module=cli
2021-12-01 09:14:23.606 cli INF renameRemoteTriggersListKeyForwards done module=cli
2021-12-01 09:14:39.584 cli INF renameTagSubscriptionsKeyForwards done module=cli
2021-12-01 09:14:55.416 cli INF renameTagTriggersKeyKeyForwards done module=cli
2021-12-01 09:14:55.416 cli INF moveToClusterForwards done module=cli

real    0m31.840s
user    0m8.401s
sys     0m12.451s
```

3. (run command reverse)
```bash
$ ./build/cli -config local/cli.yml -move-to-cluster-reverse
```

### DIFF
--- a/cmd/cli/cluster.go
+++ b/cmd/cli/cluster.go
@@ -1,0 +1,225 @@
+package main
+
+import (
+	"strings"
+
+	"github.com/moira-alert/moira"
+	"github.com/moira-alert/moira/database/redis"
+)
+
+var noSuchKeyError = "ERR no such key"
+var anyTagsSubscriptionsKeyOld = "moira-any-tags-subscriptions"
+var anyTagsSubscriptionsKeyNew = "{moira-tag-subscriptions}:moira-any-tags-subscriptions"
+var triggersListKeyOld = "moira-triggers-list"
+var triggersListKeyNew = "{moira-triggers-list}:moira-triggers-list"
+var remoteTriggersListKeyOld = "moira-remote-triggers-list"
+var remoteTriggersListKeyNew = "{moira-triggers-list}:moira-remote-triggers-list"
+var tagSubscriptionsKeyPrefixOld = "moira-tag-subscriptions:"
+var tagSubscriptionsKeyPrefixNew = "{moira-tag-subscriptions}:"
+var tagTriggersKeyKeyPrefixOld = "moira-tag-triggers:"
+var tagTriggersKeyKeyPrefixNew = "{moira-tag-triggers}:"
+
+func renameKey(database moira.Database, oldKey string, newKey string) error {
+	switch d := database.(type) {
+	case *redis.DbConnector:
+		err := d.Client().Rename(d.Context(), oldKey, newKey).Err()
+		if err != nil {
+			if err.Error() != noSuchKeyError {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+func changeKeysPrefix(database moira.Database, oldPrefix string, newPrefix string) error {
+	switch d := database.(type) {
+	case *redis.DbConnector:
+		pipe := d.Client().TxPipeline()
+		iter := d.Client().Scan(d.Context(), 0, oldPrefix+"*", 0).Iterator()
+		for iter.Next(d.Context()) {
+			oldKey := iter.Val()
+			newKey := strings.Replace(iter.Val(), oldPrefix, newPrefix, 1)
+			pipe.Rename(d.Context(), oldKey, newKey)
+		}
+		_, err := pipe.Exec(d.Context())
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func renameAnyTagsSubscriptionsKeyForwards(logger moira.Logger, database moira.Database) error {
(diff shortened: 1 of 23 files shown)


dmitryanchikov (dismissed):
LGTM! Couple questions below:
1. Are there any other keys which need to be migrated as well? Or a full list of keys is handled in this PR? Why only these?
2. Could you please add brief but complete description to PR? and how this new command should be executed.
3. Looks like there is no section in docs dedicated to CLI usage (commands, params, requirements, etc), can you check? Would it worthwhile to create a ticket for that?

cmd/cli/cluster.go
```diff
@@ -0,0 +1,158 @@
+package main
+
+import (
+	"strings"
+
+	"github.com/moira-alert/moira"
+	"github.com/moira-alert/moira/database/redis"
+)
+
+func renameAnyTagsSubscriptionsKeyForwards(database moira.Database) error {
```
KiskachiMaria, line 10: The difference between most rename methods is key strings. It looks like a good place for creating common function: `func renameKey(database moira.Database, key string, newKey string)`

cmd/cli/cluster.go
```diff
@@ -0,0 +1,158 @@
...
+	switch d := database.(type) {
+	case *redis.DbConnector:
+		err := d.Client().Rename(d.Context(), "moira-any-tags-subscriptions", "{moira-tag-subscriptions}:moira-any-tags-subscriptions").Err()
```
KiskachiMaria, line 13: For correct reversing is very important to have the same strings for keys. I suggest saving each key like a string constant to protect against misprints.

cmd/cli/cluster.go
```diff
@@ -0,0 +1,158 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+func renameAnyTagsSubscriptionsKeyReverse(database moira.Database) error {
+	switch d := database.(type) {
+	case *redis.DbConnector:
+		err := d.Client().Rename(d.Context(), "{moira-tag-subscriptions}:moira-any-tags-subscriptions", "moira-any-tags-subscriptions").Err()
```
KiskachiMaria, line 24: Rename returns error when key doesn't exist in database. It can be the reason for incorrect exit of migration process.

cmd/cli/main.go
```diff
@@ -56,6 +56,11 @@ var (
 	triggerDumpFile = flag.String("trigger-dump-file", "", "File that holds trigger dump JSON from api method response")
 )
 
+var (
```
KiskachiMaria, line 59: After that PR cli will contain two commands for updating: "update from-version" and "move-to-cluster-forwards". How the user should choose between them?
At the same time, the new database format will be used not only with cluster-redis but with any type of redis. So, the format of database keys is connected with version, but not with redis type. And I think that better include new migration command into "update from-version".

KiskachiMaria (changes requested):

zhelyabuzhsky:
@dmitryanchikov thanks for your review 🙂


> 1. Are there any other keys which need to be migrated as well? Or a full list of keys is handled in this PR? Why only these?

There are no other keys to migrate. I added some additional information to the description.

> 2. Could you please add brief but complete description to PR? and how this new command should be executed.

No problem :) Done

> 3. Looks like there is no section in docs dedicated to CLI usage (commands, params, requirements, etc), can you check? Would it worthwhile to create a ticket for that?

I will add this information to the changelog of the future release.

cmd/cli/cluster.go
zhelyabuzhsky: I added a check, thanks.

cmd/cli/main.go
zhelyabuzhsky: I agree with your remark. But at the moment, this is a script for migrating Kontur's installation. The new release of Moira and migration script will be done before merging to master.

cmd/cli/cluster.go
zhelyabuzhsky: Done

cmd/cli/cluster.go
zhelyabuzhsky: Done

KiskachiMaria (approved):

dmitryanchikov (approved):

PR #696, opened 2021-12-07
Added Redis Cluster support

zhelyabuzhsky:
The application has been rewritten from [redigo](https://github.com/gomodule/redigo) to [go-redis](https://github.com/go-redis/redis). Now it supports not only standalone Redis and Redis Sentinel, but **also** [Redis Cluster](https://redis.io/topics/cluster-spec).

### Redis Cluster goals

- High performance and linear scalability up to 1000 nodes. There are no proxies, asynchronous replication is used, and no merge operations are performed on values.
- Acceptable degree of write safety: the system tries (in a best-effort way) to retain all the writes originating from clients connected with the majority of the master nodes. Usually there are small windows where acknowledged writes can be lost. Windows to lose acknowledged writes are larger when clients are in a minority partition.
- Availability: Redis Cluster is able to survive partitions where the majority of the master nodes are reachable and there is at least one reachable replica for every master node that is no longer reachable. Moreover using replicas migration, masters no longer replicated by any replica will receive one from a master which is covered by multiple replicas.


cmd/cli/extra.go
```diff
@@ -30,3 +30,27 @@ func enablePlottingInAllSubscriptions(logger moira.Logger, database moira.Databa
 	}
 	return nil
 }
+
+func makeAllTriggersDataSourceLocal(logger moira.Logger, database moira.Database) error {
```
msaf1980, line 34: May be not needed in master ? Function will be dangerous. when running on production.

zhelyabuzhsky: done https://github.com/moira-alert/moira/pull/725

KiskachiMaria (approved):

dmitryanchikov (dismissed):

KiskachiMaria (approved):

PR #557, opened 2020-06-27
Add CodeQL analysis

beevee:
# PR Summary

GitHub offered early access for something called Code Scanning. I'm giving it a try.

PR #705, opened 2021-12-09
Added CodeQL analysis

zhelyabuzhsky:
More information about [CodeQL](https://codeql.github.com)

Related to https://github.com/moira-alert/moira/pull/557

### DIFF
--- a/checker/worker/worker.go
+++ b/checker/worker/worker.go
@@ -1,6 +1,7 @@
 package worker
 
 import (
+	"errors"
 	"runtime"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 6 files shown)


checker/worker/worker.go
```diff
@@ -67,7 +68,13 @@ func (worker *Checker) Start() error {
 		worker.Logger.Info("Remote checker disabled")
 	}
 
+	const maxParallelChecksMaxValue = 1024 * 8
```
dmitryanchikov, line 71: Can you leave a comment why is that const set exactly to that value? 
Also, since `MaxParallelChecks ` is already placed in config, does this const really useful? Wouldn't we like to leave it up to user set whatever they need?

checker/worker/worker.go
```diff
@@ -79,6 +86,11 @@ func (worker *Checker) Start() error {
 	}
 
 	if worker.remoteEnabled {
+		const maxParallelRemoteChecksMaxValue = 1024 * 8
```
dmitryanchikov, line 89: Same as above

dmitryanchikov (commented):
Looks good, couple questions below

KiskachiMaria (approved):

checker/worker/worker.go
zhelyabuzhsky: > why is that const set exactly to that value?

I used the current value from the config file with a margin

dmitryanchikov (approved):

PR #718, opened 2022-01-18
Sharded channels

zhelyabuzhsky:
In the case of Redis Cluster, which node the channel hit, it becomes very loaded. To avoid this situation, the channel was sharded. Also made some refactoring to keep logic in the `master` branch.

### DIFF
--- a/database/redis/database.go
+++ b/database/redis/database.go
@@ -132,6 +132,7 @@ func (connector *DbConnector) manageSubscriptions(tomb *tomb.Tomb, channels []st
 				connector.logger.Info("Reconnected to subscription")
 				psc = newPsc
 				<-time.After(receiveErrorSleepDuration)
+				continue
 			}
 			switch data := raw.(type) {
 			case *redis.Message:
--- a/database/redis/database.go
+++ b/database/redis/database.go
@@ -2,7 +2,8 @@ package redis
 
 import (
 	"context"
-	"net"
+	"fmt"
+	"io"
 	"time"
 
 	"github.com/go-redis/redis/v8"
@@ -91,9 +92,17 @@ func NewTestDatabaseWithIncorrectConfig(logger moira.Logger) *DbConnector {
 	return NewDatabase(logger, Config{Addrs: []string{"0.0.0.0:0000"}}, testSource)
 }
 
+func (connector *DbConnector) makePubSubConnection(channels []string) (*redis.PubSub, error) {
+	psc := (*connector.client).Subscribe(connector.context, channels...)
+	err := psc.Ping(connector.context)
+	if err != nil {
+		return nil, fmt.Errorf("failed to subscribe to '%s', error: %v", channels, err)
+	}
+	return psc, nil
+}
+
 func (connector *DbConnector) manageSubscriptions(tomb *tomb.Tomb, channels []string) (<-chan []byte, error) {
-	c := (*connector.client).Subscribe(connector.context, channels...)
-	err := c.Ping(connector.context)
+	psc, err := connector.makePubSubConnection(channels)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


database/redis/metric.go
```diff
@@ -264,7 +268,13 @@ func (connector *DbConnector) needRemoveMetrics(metric string) bool {
 }
 
 var patternsListKey = "moira-pattern-list"
-var metricEventKey = "{channel}:metric-event"
+
```
msaf1980, line 271: As a variant - use parameter like `event_sharded` (boolean). Also number of shards can be used (but need to sync beetween and notifiers).

dmitryanchikov (dismissed):
LGTM!

database/redis/metric.go
zhelyabuzhsky: Thanks for your suggestion, but I think it's premature.

database/redis/database.go
```diff
@@ -91,23 +92,47 @@ func NewTestDatabaseWithIncorrectConfig(logger moira.Logger) *DbConnector {
 	return NewDatabase(logger, Config{Addrs: []string{"0.0.0.0:0000"}}, testSource)
 }
 
-func (connector *DbConnector) manageSubscriptions(tomb *tomb.Tomb, channel string) (<-chan []byte, error) {
-	c := (*connector.client).Subscribe(connector.context, channel)
-	err := c.Ping(connector.context)
+func (connector *DbConnector) makePubSubConnection(channels []string) (*redis.PubSub, error) {
+	psc := (*connector.client).Subscribe(connector.context, channels...)
```
dmitryanchikov, line 96: Personally I vote for readable variable names, like `pubSubChannel` in this case and avoid using one character variable names

database/redis/database.go
```diff
@@ -91,23 +92,47 @@ func NewTestDatabaseWithIncorrectConfig(logger moira.Logger) *DbConnector {
...
+				connector.logger.Infof("Receive() returned error: %s. Reconnecting...", err.Error())
+				newPsc, err := connector.makePubSubConnection(channels)
+				if err != nil {
+					connector.logger.Errorf("Failed to reconnect to subscription: %v", err)
+					<-time.After(receiveErrorSleepDuration)
+					continue
+				}
+				connector.logger.Info("Reconnected to subscription")
+				psc = newPsc
+				<-time.After(receiveErrorSleepDuration)
```
dmitryanchikov, line 134: Maybe extract this peace of code to separate function to make it more readable?

database/redis/database.go
```diff
@@ -91,23 +92,47 @@ func NewTestDatabaseWithIncorrectConfig(logger moira.Logger) *DbConnector {
...
+				}
+				connector.logger.Infof("Receive() returned error: %s. Reconnecting...", err.Error())
+				newPsc, err := connector.makePubSubConnection(channels)
+				if err != nil {
+					connector.logger.Errorf("Failed to reconnect to subscription: %v", err)
+					<-time.After(receiveErrorSleepDuration)
+					continue
+				}
+				connector.logger.Info("Reconnected to subscription")
+				psc = newPsc
```
dmitryanchikov, line 133: Couple questions here: 
- will this variable be picked up by gouroutine with unsubscribing on 112-113 lines?
- shouldn't we unsubscribe first from previous object?

dmitryanchikov (commented):
Coulpe more questions below

database/redis/metric.go
```diff
@@ -128,7 +130,9 @@ func (connector *DbConnector) SaveMetrics(metrics map[string]*moira.MatchedMetri
 				continue
 			}
 
-			if err = c.Publish(ctx, metricEventKey, event).Err(); err != nil {
+			rand.Seed(time.Now().UnixNano())
```
KiskachiMaria, line 133: Usually getting time is an expensive operation. I suggest doing it one time.

database/redis/database.go
zhelyabuzhsky: I totally agree with the idea of giving more readable names to variables, but if I do it now, the diff of the epic branch will increase, and it is already huge.

database/redis/database.go
zhelyabuzhsky: I totally agree with your proposal
> but if I do it now, the diff of the epic branch will increase, and it is already huge.

database/redis/database.go
zhelyabuzhsky: > will this variable be picked up by gouroutine with unsubscribing on 112-113 lines?

> shouldn't we unsubscribe first from previous object?

I really don't know :) There is a complete repetition of the logic of the master branch.

database/redis/metric.go
zhelyabuzhsky: > I suggest doing it one time.

How exactly do you propose to do this?

KiskachiMaria: I suggest moving the line `rand.Seed(time.Now().UnixNano())` before line `for _, metric := range metrics {`.

zhelyabuzhsky: Done

KiskachiMaria (approved):

dmitryanchikov (approved):

PR #723, opened 2022-02-01
Improved GetAllContacts method

zhelyabuzhsky:
before:

> GET http://127.0.0.1:8080/api/contact HTTP/1.1 from 127.0.0.1:39580 by anonymous - 200 155663B in **1m53.036403899s**

after:

> GET http://127.0.0.1:8080/api/contact HTTP/1.1 from 127.0.0.1:39580 by anonymous - 200 155663B in **2.671802051s**

### DIFF
--- a/database/redis/contact.go
+++ b/database/redis/contact.go
@@ -58,6 +58,29 @@ func (connector *DbConnector) GetContacts(contactIDs []string) ([]*moira.Contact
 	return contacts, nil
 }
 
+func getContactsKeysOnRedisNode(ctx context.Context, client redis.UniversalClient) ([]string, error) {
+	var cursor uint64
+	var keys []string
+	const scanCount = 10000
+
+	for {
+		var keysResult []string
+		var err error
+		keysResult, cursor, err = client.Scan(ctx, cursor, contactKey("*"), scanCount).Result()
+
+		if err != nil {
+			return nil, err
+		}
+
+		keys = append(keys, keysResult...)
+
+		if cursor == 0 {
+			break
+		}
+	}
+	return keys, nil
+}
+
 // GetAllContacts returns full contact list
 func (connector *DbConnector) GetAllContacts() ([]*moira.ContactData, error) {
 	client := *connector.client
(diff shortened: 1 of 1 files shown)


database/redis/contact.go
```diff
@@ -66,22 +66,44 @@ func (connector *DbConnector) GetAllContacts() ([]*moira.ContactData, error) {
...
+
+			if err != nil {
+				return nil, err
+			}
+
+			keys = append(keys, keysResult...)
+
+			if cursor == 0 {
+				break
+			}
```
dmitryanchikov, line 106: Please extract dublicate code into separate function

database/redis/contact.go
```diff
@@ -66,22 +66,44 @@ func (connector *DbConnector) GetAllContacts() ([]*moira.ContactData, error) {
...
 	case *redis.ClusterClient:
 		err := c.ForEachMaster(connector.context, func(ctx context.Context, shard *redis.Client) error {
-			iter := shard.Scan(ctx, 0, contactKey("*"), 0).Iterator()
-			for iter.Next(ctx) {
-				keys = append(keys, iter.Val())
+			var cursor uint64
+			var err error
+			var keysResult []string
+			for {
+				keysResult, cursor, err = shard.Scan(connector.context, cursor, contactKey("*"), 10000).Result()
```
dmitryanchikov, line 73: Maybe create function level const for 10000?

dmitryanchikov (changes requested):
Nice work overall! A few suggestions are below:

database/redis/contact.go
zhelyabuzhsky: It's not as easy as it seems)

Сan you show a specific working solution?

database/redis/contact.go
dmitryanchikov: Please double-check if there should be `ctx` instead of `connector.context` on that line

database/redis/contact.go
dmitryanchikov: Sure, I gave it a shot and here are my findings. Below is the function exculding duplicates:
```
func scanContactKeysFromNode(ctx context.Context, client redis.UniversalClient) ([]string, error) {
	var cursor uint64
	var keys []string
	const scanCount = 10000
	for {
		keysResult, cursor, err := client.Scan(ctx, cursor, contactKey("*"), scanCount).Result()

		if err != nil {
			return nil, err
		}

		keys = append(keys, keysResult...)

		if cursor == 0 {
			break
		}
	}
	return keys, nil
}
```

And here is updated `GetAllContacts` function:
```
// GetAllContacts returns full contact list
func (connector *DbConnector) GetAllContacts() ([]*moira.ContactData, error) {
	client := *connector.client
	var keys []string

	switch c := client.(type) {
	case *redis.ClusterClient:
		err := c.ForEachMaster(connector.context, func(ctx context.Context, shard *redis.Client) error {
			keysResult, err := scanContactKeysFromNode(ctx, shard)
			if err != nil {
				return err
			}
			keys = append(keys, keysResult...)
			return nil
		})

		if err != nil {
			return nil, err
		}
	default:
		keysResult, err := scanContactKeysFromNode(connector.context, c)
		if err != nil {
			return nil, err
		}
		keys = append(keys, keysResult...)
	}

	contactIDs := make([]string, 0, len(keys))
	for _, key := range keys {
		key = key[14:]
		contactIDs = append(contactIDs, key)
	}
	return connector.GetContacts(contactIDs)
}
```

database/redis/contact.go
zhelyabuzhsky: Done

database/redis/contact.go
zhelyabuzhsky: Thank you :) Done

dmitryanchikov (approved):
👍 Thanks for updates!

PR #742, opened 2022-04-15
Added wildcard support to SeriesByTag target

zhelyabuzhsky:
Wildcards like `responseCode={405,406,407,411,413,414,415}` were processed incorrectly before. This PR fixes this problem.


dmitryanchikov (approved):
lgtm

PR #753, opened 2022-06-02
Fixed abandoned trigger last checks

zhelyabuzhsky:
In case of an error during deletion of trigger, its last check were left unremoved. This led not only to clogging the database, but also to problems when creating a trigger with the same identifier.

In this PR:
- added command to clean up the database from abandoned lastchecks
- the section of the code has been rewritten so that the removal of the trigger takes place in one transaction with the removal of its last check


dmitryanchikov (commented):
Overall looks good, I might miss but are there any new tests for checking of last check removing on trigger removing?

database/redis/last_check.go
```diff
@@ -71,14 +75,20 @@ func (connector *DbConnector) getSelfStateCheckCountKey(isRemote bool) string {
 	return selfStateChecksCounterKey
 }
 
+func appendRemoveTriggerLastCheckToRedisPipeline(context context.Context, pipe redis.Pipeliner, triggerID string) redis.Pipeliner {
```
kissken, line 78: optional: Variable name is same as package name. Maybe better change to "ctx"?

kissken (commented):
Left a small cosmetic comment.

database/redis/last_check.go
zhelyabuzhsky: Changed to `ctx`

zhelyabuzhsky:
> are there any new tests for checking of last check removing on trigger removing?

[Added](https://github.com/moira-alert/moira/pull/753/files#diff-ecefb56af45b7a8512d30dba0fcbb400176762b1464abaa97f518d4fed393536R288-R291) assertion to test case, thanks )

dmitryanchikov (approved):

kissken (approved):

PR #761, opened 2022-06-27
Fixed error handling when getting metrics

kissken:
1. When carbon-api returns "no timeseries with that name", it [concatenates additional text into error](https://github.com/go-graphite/carbonapi/blob/main/expr/expr.go#L162). So we cannot compare error as "equal"
2. When  carbon-api returns "ErrUnknownFunction", it [wraps error with "merry.WithHTTPCode"](https://github.com/go-graphite/carbonapi/blob/main/expr/expr.go#L181). So, we need Unwrap error before compare it with helper.ErrUnknownFunction.

### DIFF
--- a/metric_source/local/local.go
+++ b/metric_source/local/local.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"runtime/debug"
+	"strings"
 
 	"github.com/go-graphite/carbonapi/expr"
 	"github.com/go-graphite/carbonapi/expr/functions"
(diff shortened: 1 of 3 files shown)


metric_source/local/local_test.go
```diff
@@ -88,7 +89,7 @@ func TestEvaluateTarget(t *testing.T) {
 			dataBase.EXPECT().GetMetricsValues([]string{metric}, from, until).Return(dataList, nil)
 			dataBase.EXPECT().GetMetricsTTLSeconds().Return(metricsTTL)
 			result, err := localSource.Fetch("aliasByNoe(super.puper.pattern, 2)", from, until, true)
-			So(err.Error(), ShouldResemble, "failed to evaluate target 'aliasByNoe(super.puper.pattern, 2)': unknown function in evalExpr: \"aliasByNoe\"")
+			So(err.Error(), ShouldResemble, "unknown function in evalExpr: \"aliasByNoe\"")
```
zhelyabuzhsky, line 92: ?

metric_source/local/local_test.go
```diff
@@ -322,3 +323,44 @@ func TestLocal_IsConfigured(t *testing.T) {
 		So(actual, ShouldBeTrue)
 	})
 }
+
+func TestLocal_getMetricData(t *testing.T) {
+	Convey("When all is correct, we dont return any error", t, func() {
```
zhelyabuzhsky, line 328:
```suggestion
	Convey("When everything is correct, we don't return any error", t, func() {
```

metric_source/local/local.go
```diff
@@ -132,3 +118,29 @@ func getPatternsMetricData(database moira.Database, patterns []parser.MetricRequ
...
+		}
+	}()
+
+	result, err = expr.EvalExpr(context.Background(), expression, from, until, metricsMap)
+	if err != nil {
+		switch {
+		case strings.Contains(err.Error(), parser.ErrSeriesDoesNotExist.Error()):
+			break
+		case isErrUnknownFunction(err):
+			return nil, err
```
zhelyabuzhsky, line 136:
```suggestion
      return nil, ErrorUnknownFunction(err)
```

zhelyabuzhsky (changes requested):
Norm, no....

kissken:
needs to open, when we will be update carbon-api

PR #762, opened 2022-06-27
Updated github.com/go-graphite/carbonapi to v0.15.2

kissken:


kissken:
will be open when we update carbon-api again

PR #772, opened 2022-07-14
Added cleanup abandoned tags

Dimedrolity:
Abandoned tags are tags without triggers, so it is a trash in database and affects users in Web UI, these tags should be removed. Now tags would be removed on cleanup.


Dimedrolity:
PR has been closed, because rename git branch should be renamed. New PR with renamed branch [here ](https://github.com/moira-alert/moira/pull/773) @zhelyabuzhsky

PR #773, opened 2022-07-15
Added cleanup abandoned tags

Dimedrolity:
Abandoned tags are tags without triggers and subscriptions, so it is a trash in database and affects users in Web UI, these tags should be removed. Now tags would be removed on cleanup.

### DIFF
--- a/cmd/cli/main.go
+++ b/cmd/cli/main.go
@@ -177,6 +177,13 @@ func main() { //nolint
 			log.Error(err)
 		}
 		log.Info("Cleanup abandoned triggers last checks finished")
+
+		log.Info("Cleanup abandoned tags started")
+		count, err := handleCleanUpAbandonedTags(dataBase)
+		if err != nil {
+			log.Error(err)
+		}
+		log.Info("Cleanup abandoned tags finished. ", fmt.Sprintf("Count of deleted tags is %d", count))
 	}
 
 	if *cleanupRetentions {
--- a/database/redis/tag.go
+++ b/database/redis/tag.go
@@ -56,20 +56,20 @@ func (connector *DbConnector) CleanUpAbandonedTags() (int, error) {
 
 		result, err := client.Exists(connector.context, tagTriggersKey(tag)).Result()
 		if err != nil {
-			return 0, fmt.Errorf("failed to check tag triggers existence, error: %v", err)
+			return count, fmt.Errorf("failed to check tag triggers existence, error: %v", err)
 		}
 		isTriggerExists := result == 1
 
 		result, err = client.Exists(connector.context, tagSubscriptionKey(tag)).Result()
 		if err != nil {
-			return 0, fmt.Errorf("failed to check tag subscription existence, error: %v", err)
+			return count, fmt.Errorf("failed to check tag subscription existence, error: %v", err)
 		}
 		isSubscriptionExists := result == 1
 
 		if !isTriggerExists && !isSubscriptionExists {
 			err = client.SRem(connector.context, tagsKey, tag).Err()
 			if err != nil {
-				return 0, err
+				return count, err
 			}
 
 			count++
--- a/database/redis/tag.go
+++ b/database/redis/tag.go
@@ -58,22 +58,23 @@ func (connector *DbConnector) CleanUpAbandonedTags() (int, error) {
 		if err != nil {
 			return count, fmt.Errorf("failed to check tag triggers existence, error: %v", err)
 		}
-		isTriggerExists := result == 1
+		if isTriggerExists := result == 1; isTriggerExists {
+			continue
+		}
 
 		result, err = client.Exists(connector.context, tagSubscriptionKey(tag)).Result()
 		if err != nil {
 			return count, fmt.Errorf("failed to check tag subscription existence, error: %v", err)
 		}
-		isSubscriptionExists := result == 1
-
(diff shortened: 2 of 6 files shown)


Dimedrolity:
@zhelyabuzhsky I removed cluster specific code, check it, please

cmd/cli/main.go
```diff
@@ -175,6 +175,14 @@ func main() { //nolint
 			log.Error(err)
 		}
 		log.Info("Cleanup abandoned triggers last checks finished")
+
+		log.Info("Cleanup abandoned tags started")
+		count, err := handleCleanUpAbandonedTags(dataBase)
+		if err != nil {
+			log.Info(fmt.Sprintf("Count of deleted tags is %d", count))
```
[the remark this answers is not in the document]
Dimedrolity: thanks, fixed

kissken (dismissed):

database/redis/tag.go
```diff
@@ -44,6 +44,41 @@ func (connector *DbConnector) GetTagTriggerIDs(tagName string) ([]string, error)
...
+	var count int
+	client := *connector.client
+
+	iter := client.SScan(connector.context, tagsKey, 0, "*", 0).Iterator()
+	for iter.Next(connector.context) {
+		tag := iter.Val()
+
+		result, err := client.Exists(connector.context, tagTriggersKey(tag)).Result()
+		if err != nil {
+			return 0, fmt.Errorf("failed to check tag triggers existence, error: %v", err)
```
Tetrergeru, line 59: If we fail to check existance of one tag, we stop the whole cleanup? Maybe we should just skip this tag and then somehow return the list of erroneous tags?

Tetrergeru: We return 0, even though we might have deleted some tags, if I understand the code correctly

database/redis/tag.go
```diff
@@ -44,6 +44,41 @@ func (connector *DbConnector) GetTagTriggerIDs(tagName string) ([]string, error)
...
+		}
+		isTriggerExists := result == 1
```
Tetrergeru, line 61:
```suggestion
		if result == 1 { 
		     continue;
		}
```
We don't have to check existance of subscription if the trigger exists, right?

Tetrergeru (changes requested):
Left some comments

database/redis/tag.go
Dimedrolity: > If we fail to check existance of one tag, we stop the whole cleanup?

Yes. We need to cleanup all abandoned tags. If there is an error, we should fix it and then cleanup all abandoned tags. 

> Maybe we should just skip this tag and then somehow return the list of erroneous tags?
It is nice approach, but I think here it is overhead

Dimedrolity: > We return 0, even though we might have deleted some tags, if I understand the code correctly

Thanks, fixed https://github.com/moira-alert/moira/pull/773/commits/e0cf2e633c291cd7c883ae1e7ccb9bacf91bb685

database/redis/tag.go
Dimedrolity: It is fine. I don't follow premature optimization. That code executes pretty fast on our data

Tetrergeru: I think, we should use continue here primarily to improve redability

Tetrergeru: Maybe it is a subjective thing. If you think that it would make code more complicated -- let's keep it the way it is

Dimedrolity: 🤔  But do we need to improve readability? I think this code is simple enough

Tetrergeru: In my opinion, it is almost always worth it to improve readablility

Dimedrolity: To be precise I think you mean `continue` here is not for readability, but for better understanding. Current code is understandable enough. I try to not to be perfectionist. Also I think cli service in Moira is less important than other, it is kind a helper, so we don't need pay to much attention to it

Tetrergeru: Sorry, we unfortunately spend more time on this debate, than we would on applying these changes to the code. I don't really understand why do you resist so much :(

Dimedrolity: > Sorry, we unfortunately spend more time on this debate, than we would on applying these changes to the code.

I agree with you. 

>  I don't really understand why do you resist so much :(

Because I don't see any profit of suggested change

Dimedrolity: @Tetrergeru I fixed it, check it please https://github.com/moira-alert/moira/pull/773/commits/4554380c75d807618ed4eb9be65a42aaa2899a4f

Tetrergeru (approved):

cmd/cli/main.go
```diff
@@ -177,6 +177,13 @@ func main() { //nolint
 			log.Error(err)
 		}
 		log.Info("Cleanup abandoned triggers last checks finished")
+
+		log.Info("Cleanup abandoned tags started")
+		count, err := handleCleanUpAbandonedTags(dataBase)
+		if err != nil {
+			log.Error(err)
+		}
+		log.Info("Cleanup abandoned tags finished. ", fmt.Sprintf("Count of deleted tags is %d", count))
```
kissken, line 186: @Tetrergeru seems u have to change ur pr, don't forget

kissken (approved):

PR #774, opened 2022-07-21
Fixed data races

Dimedrolity:
There were tests with data races - `TestIndex_CreateAndFill` and `TestSelfCheckWorker_selfStateChecker`. I fixed races and add `-race` flag to Makefile and ".run.xml"

### DIFF
--- a/cmd/notifier/main.go
+++ b/cmd/notifier/main.go
@@ -100,7 +100,7 @@ func main() {
 	selfState := selfstate.NewSelfCheckWorker(logger, database, sender, config.Notifier.SelfState.getSettings())
 	if err := selfState.Start(); err != nil {
 		if errors.Is(err, selfstate.ErrDisabled) {
-			logger.Warning(err.Error())
+			logger.Info(err.Error())
 		} else {
 			logger.Fatalf("Failed to start self check worker: %v", err)
 		}
--- a/cmd/notifier/main.go
+++ b/cmd/notifier/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"flag"
 	"fmt"
 	"os"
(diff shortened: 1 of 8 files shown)


notifier/selfstate/selfstate.go
```diff
@@ -24,19 +28,25 @@ type SelfCheckWorker struct {
 	Notifier   notifier.Notifier
 	Config     Config
 	tomb       tomb.Tomb
-	Heartbeats []heartbeat.Heartbeater
+	heartbeats []heartbeat.Heartbeater
+}
+
+func NewSelfCheckWorker(logger moira.Logger, database moira.Database, notifier notifier.Notifier, config Config) *SelfCheckWorker {
```
kissken, line 34: missed godoc and maybe we change order of variables? 

example: config first and logger last?

notifier/selfstate/selfstate.go
```diff
@@ -24,19 +28,25 @@ type SelfCheckWorker struct {
...
-		return nil
+		selfCheck.Logger.Debugf(ErrDisabled.Error())
+		return ErrDisabled
 	}
+
 	senders := selfCheck.Notifier.GetSenders()
 	if err := selfCheck.Config.checkConfig(senders); err != nil {
-		selfCheck.Logger.Errorf("Can't configure Moira Self State Monitoring: %s", err.Error())
-		return nil
+		selfCheck.Logger.Errorf(ErrWrongConfig.Error())
```
kissken, line 48: Will be logged twice?

kissken (changes requested):
cool 💪

notifier/selfstate/selfstate.go
Dimedrolity: yes, thanks. Fixed here https://github.com/moira-alert/moira/pull/774/commits/e44f41735e609b58dd48bb95a35008b101f7af5c

notifier/selfstate/selfstate.go
Dimedrolity: I think godoc is not necessary, because it is constructor func (`New` prefix is a naming convention). We can write something about heartbeats, but idk.

Order of variables. I think it is ok to start with logger and db like `notifier.NewNotifier(database, logger, config, ...)` or `handler.NewHandler(database, logger, ...)` because we use logger and db everywhere and it is good to place it first like context.Context. It is more easy to add new parameters to the end

cmd/notifier/main.go
```diff
@@ -148,6 +148,10 @@ func stopNotificationsFetcher(worker *notifications.FetchNotificationsWorker) {
 
 func stopSelfStateChecker(checker *selfstate.SelfCheckWorker) {
 	if err := checker.Stop(); err != nil {
-		logger.Errorf("Failed to stop self check worker: %v", err)
+		if errors.Is(err, selfstate.ErrDisabled) {
+			logger.Warning(err.Error())
```
kissken, line 152: oops :) 

lets make logger.Warning(err)

Dimedrolity: There are `logger.Warning` and `Errorf` inside if statement. I think it is enough

kissken: no, in first you use err.Error() (however logger.Warning use interface)

and second use "err"

lets make it the same type

notifier/selfstate/selfstate.go
kissken: by golang code-style all public func have to get godoc

PR #778, opened 2022-07-27
Refactored cluster specific code

Dimedrolity:
## DRY (Don't Repeat Yourself)
There was repeat of switch-case for call function on Redis cluster or standalone. Extracted method to call functions dependent of Redis client type.


Dimedrolity:
Should I test this new method?

database/redis/database.go
```diff
@@ -203,3 +196,22 @@ func (connector *DbConnector) Client() redis.UniversalClient {
 func (connector *DbConnector) Context() context.Context {
 	return connector.context
 }
+
+// callDependent calls the fn dependent of Redis client type (cluster or standalone).
+func (connector *DbConnector) callDependent(fn func(connector *DbConnector, client redis.UniversalClient) error) error {
```
kissken, line 201: maybe named better: callFunc? 

verb + adjective seems not good :)

kissken: and can we write a test on this small func?

Dimedrolity: renamed here https://github.com/moira-alert/moira/pull/778/commits/1f59f98dd5964a1ec452cf468e0385620912e3f1.

This func is internal, so it is already (partially) covered by tests and tests are passed. I was thinking about test, and we can test that function `fn` was called once or 3 times, and that test will depend on current Redis client type, cluster or standalone. I think unit test should not depend on current Redis client type (it should be private details). And it is problem that we cannot fully test this function, it is covered only by half of it.

Dimedrolity:
Converted to draft because we can't test "cli" service right now

kissken:
please, let's fix commits

Dimedrolity:
@kissken Please, say more specific, what you suggest to fix?

kissken:
@Dimedrolity fix this :)
image

Dimedrolity:
@kissken Excuse me, I didn't get what you mean. What is mean "fix commit"? Is there a bug in commits that I should fix?

kissken:
@Dimedrolity first commit message is not pretty, let's make it good or let's concatenate two in one )

Dimedrolity:
@kissken Ok, I squashed it

kissken (dismissed):

kissken (dismissed):

database/redis/database.go
```diff
@@ -124,3 +117,22 @@ func (connector *DbConnector) Client() redis.UniversalClient {
...
+func (connector *DbConnector) callFunc(fn func(connector *DbConnector, client redis.UniversalClient) error) error {
+	client := *connector.client
+
+	switch c := client.(type) {
+	case *redis.ClusterClient:
+		return c.ForEachMaster(connector.context, func(ctx context.Context, shard *redis.Client) error {
+			return fn(connector, shard)
+		})
+	default:
+		err := fn(connector, client)
```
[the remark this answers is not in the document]
almostinf: Fixed

kissken (approved):

PR #783, opened 2022-08-11
Fixed error handle

Dimedrolity:
If there is no data for metric, carbonapi consider it error `parser.ErrSeriesDoesNotExist`, but for Moira it is ok! It can happen when Graphite stopped or Moira clients send events instead of metrics.

There was wrapped error `parser.ErrSeriesDoesNotExist` that we don't handle, so some triggers gone in Exception state on cleanup of cli. Added handle.

### DIFF
--- a/metric_source/local/local.go
+++ b/metric_source/local/local.go
@@ -2,6 +2,7 @@ package local
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"runtime/debug"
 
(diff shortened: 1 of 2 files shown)


metric_source/local/local.go
```diff
@@ -98,15 +78,40 @@ func (local *Local) Fetch(target string, from int64, until int64, allowRealTimeA
...
+	defer func() {
+		if r := recover(); r != nil {
+			result = nil
+			err = ErrEvaluateTargetFailedWithPanic{target: target, recoverMessage: r, stackRecord: debug.Stack()}
+		}
+	}()
+	result, err = expr.EvalExpr(context.Background(), expression, from, until, metricsMap)
+	if err != nil {
+		if errors.Is(err, parser.ErrSeriesDoesNotExist) {
+			// It is ok for us because some clients of Moira send events instead of metrics
```
kissken, line 100: let's remove it

metric_source/local/local_test.go
```diff
@@ -322,3 +324,47 @@ func TestLocal_IsConfigured(t *testing.T) {
...
+		expression, _, err := parser.ParseExpr(`alias( divideSeries( alias( sumSeries( exclude( groupByNode( OFD.Production.{ofd-api,ofd-front}.*.fns-service-client.v120.*.GetCashboxRegistrationInformationAsync.ResponseCode.*.Meter.Rate-15-min-Requests-per-s, 9, "sum" ), "Ok" ) ), "bad" ), alias( sumSeries( OFD.Production.{ofd-api,ofd-front}.*.fns-service-client.v120.*.GetCashboxRegistrationInformationAsync.ResponseCode.*.Meter.Rate-15-min-Requests-per-s ), "total" ) ), "Result" )`)
+		So(err, ShouldBeNil)
+		res, err := evalExpr("target", expression, time.Now().Add(-1*time.Hour).Unix(), time.Now().Unix(), make(map[parser.MetricRequest][]*types.MetricData))
+		So(err, ShouldBeNil)
+		So(res, ShouldBeNil)
+	})
+
+	// Do we need 2 tests below?
+
+	//Convey("When give unknown func, return it", t, func() {
```
kissken, line 354: ooops

kissken (changes requested):

metric_source/local/local.go
Dimedrolity: do you mean remove comment? You are highlighted 97-100 lines

kissken: only comment

metric_source/local/local_test.go
Dimedrolity: So do we need 2 tests below?

kissken: why not?

kissken:
see that linter doesn't pass :)

metric_source/local/local.go
Dimedrolity: why? :)

Dimedrolity: moved description to commit message and PR body

metric_source/local/local_test.go
Dimedrolity: Now I uncommented and fixed 1st test, but removed 2nd test, because there is no internal error on our version of carbonapi

kissken: ok, let's do it

kissken (approved):

PR #785, opened 2022-08-15
Added cleanup for abandoned patterns

Dimedrolity:
Pattern from moira-pattern-list key has no sense if pattern-metrics key doesn't exist, so that patterns are trash in database. Now patterns would be removed on cleanup.

**Don't do this**, see [comment](https://github.com/moira-alert/moira/pull/785#issuecomment-1221920426)


kissken:
Can we add tests on errors getting?

Dimedrolity:
@kissken We can't test negative cases because we use real Redis for tests, not stub. To use stub for Redis client we need to remove asterisk [here](https://github.com/moira-alert/moira/blob/master/database/redis/database.go#L45) and refactor. In future we will do it and test negative cases

Dimedrolity:
Don't delete from moira-pattern-list because it is used by filter for matching metrics and save to DB

kissken (approved):

PR #789, opened 2022-09-07
Add functional of creating trigger with target with flat metric

Dimedrolity:
There is a bug - can't create trigger with target with only series (without Graphite-function). Need to test method TargetVerification in current state, then add test to reproduce that bug and fix it.

### DIFF
--- a/api/dto/target.go
+++ b/api/dto/target.go
@@ -113,6 +113,7 @@ type TreeOfProblems struct {
 	TreeOfProblems *problemOfTarget `json:"tree_of_problems,omitempty"`
 }
 
+// TargetVerification validates trigger targets.
 func TargetVerification(targets []string, ttl time.Duration, isRemote bool) []TreeOfProblems {
 	functionsOfTargets := make([]TreeOfProblems, 0)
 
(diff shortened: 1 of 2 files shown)


api/dto/target.go
```diff
@@ -157,11 +163,11 @@ func checkExpression(expression parser.Expr, ttl time.Duration, isRemote bool) *
 			continue
 		}
 
-		if badFunc := checkExpression(argument, ttl, isRemote); badFunc != nil {
+		if badFunc := checkFuncExpression(argument, ttl, isRemote); badFunc != nil {
```
kissken, line 166: "checkFuncExpression" seems salt is salty

we check parsed expression, yea? and lets leave old name

api/dto/target.go
```diff
@@ -126,20 +127,25 @@ func TargetVerification(targets []string, ttl time.Duration, isRemote bool) []Tr
 			continue
 		}
 
-		functionsOfTarget.TreeOfProblems = checkExpression(expr, ttl, isRemote)
+		if expr.IsFunc() {
```
kissken, line 130: i think this "if" is part of checkExpression

kissken (changes requested):
left some comments

api/dto/target.go
Dimedrolity: I can move condition to checkExpression, and it would be like 
```go
	if !expression.IsFunc() {
		return nil
	}
```
at the start of function body. 
Do you think it would be better?

api/dto/target.go
Dimedrolity: ok, I leave old name, because `if` will be in `checkExpression`, so name `checkFuncExpression` has no sense

Dimedrolity:
@kissken I am done, also squashed it to one commit

api/dto/target.go
```diff
@@ -133,13 +134,18 @@ func TargetVerification(targets []string, ttl time.Duration, isRemote bool) []Tr
 	return functionsOfTargets
 }
 
+// checkExpression validates expression.
 func checkExpression(expression parser.Expr, ttl time.Duration, isRemote bool) *problemOfTarget {
```
kissken, line 138: this is private func. godoc is not required :)

kissken (approved):

PR #790, opened 2022-09-27
Added new cmd to remove triggers

Dimedrolity:
There is a problem - one of teams of Moira users have unused triggers. So it is a trash in database. Added command to remove triggers which have ID starting with prefix.

### DIFF
--- a/cmd/cli/triggers.go
+++ b/cmd/cli/triggers.go
@@ -1,0 +1,34 @@
+package main
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/moira-alert/moira"
+)
+
+func handleRemoveTriggersStartWith(logger moira.Logger, database moira.Database, prefix string) error {
+	triggers, err := database.GetTriggerIDsStartWith(prefix)
+	if err != nil {
+		return fmt.Errorf("can't get trigger IDs start with prefix %s: %w", prefix, err)
+	}
+
+	// Added delay because command is potentially dangerous and can delete unwanted triggers
+	const delay = 10 * time.Second
+	logger.Infof("%d triggers start with %s would be removed after %s", len(triggers), prefix, delay)
+	logger.Info("You can cancel execution by Ctrl+C")
+	time.Sleep(delay)
+
+	logger.Infof("Removing triggers start with prefix %s has started", prefix)
+	for _, id := range triggers {
+		err := database.RemoveTrigger(id)
+		if err != nil {
+			return fmt.Errorf("can't remove trigger with id %s: %w", id, err)
+		}
+	}
+	logger.Infof("Removing triggers start with prefix %s has finished", prefix)
+	logger.Infof("Count of deleted is %d", len(triggers))
+	logger.Infof("Removed triggers: %s", triggers)
+
+	return nil
+}
--- a/database/redis/trigger_test.go
+++ b/database/redis/trigger_test.go
@@ -25,7 +25,7 @@ func TestTriggerStoring(t *testing.T) {
 
 	Convey("Trigger manipulation", t, func() {
 		Convey("Test trigger has subscriptions with AnyTag is true", func() {
-			trigger := &triggers[0]
+			trigger := &testTriggers[0]
 			subscription := moira.SubscriptionData{
 				ID:                "subscriptionID-00000000000001",
 				Enabled:           true,
@@ -52,7 +52,7 @@ func TestTriggerStoring(t *testing.T) {
 		})
 
 		Convey("Test save-get-remove", func() {
-			trigger := &triggers[0]
+			trigger := &testTriggers[0]
 
 			//Check for not existing not written trigger
 			actual, err := dataBase.GetTrigger(trigger.ID)
@@ -99,7 +99,7 @@ func TestTriggerStoring(t *testing.T) {
 
 			//Now just add tag and pattern in trigger and save it
 			trigger = nil
-			changedTrigger := &triggers[1]
+			changedTrigger := &testTriggers[1]
 			err = dataBase.SaveTrigger(changedTrigger.ID, changedTrigger)
 			So(err, ShouldBeNil)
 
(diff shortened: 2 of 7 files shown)


database/redis/trigger_test.go
```diff
@@ -487,7 +487,7 @@ func TestTriggerStoring(t *testing.T) {
 			So(actual, ShouldResemble, []string{trigger.ID})
 
 			// Now update trigger
-			trigger = &triggers[1]
+			trigger = &testTriggers[1]
```
kissken, line 490: why?

cmd/cli/triggers.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+func handleRemoveTriggersByPrefix(logger moira.Logger, database moira.Database, prefix string) error {
+	triggers, err := database.GetTriggerIDsByPrefix(prefix)
+	if err != nil {
+		return err
+	}
+
+	// Added delay because command is potentially dangerous and can delete unwanted triggers
+	const delay = 10 * time.Second
```
kissken, line 16: let's define outside of the function and left comment there

cmd/cli/triggers.go
```diff
@@ -0,0 +1,33 @@
...
+	logger.Infof("%d triggers by prefix %s would be removed after %s", len(triggers), prefix, delay)
+	logger.Info("You can cancel execution by Ctrl+C")
+	time.Sleep(delay)
+
+	logger.Infof("Removing triggers by prefix %s started", prefix)
+	for _, id := range triggers {
```
kissken, line 22: pipe?

cmd/cli/triggers.go
```diff
@@ -0,0 +1,33 @@
+package main
+
+import (
+	"time"
+
+	"github.com/moira-alert/moira"
+)
+
+func handleRemoveTriggersByPrefix(logger moira.Logger, database moira.Database, prefix string) error {
+	triggers, err := database.GetTriggerIDsByPrefix(prefix)
```
kissken, line 10: does func return triggers by prefix in id, doesnt it?

cmd/cli/triggers.go
```diff
@@ -0,0 +1,33 @@
...
+	const delay = 10 * time.Second
+	logger.Infof("%d triggers by prefix %s would be removed after %s", len(triggers), prefix, delay)
+	logger.Info("You can cancel execution by Ctrl+C")
+	time.Sleep(delay)
+
+	logger.Infof("Removing triggers by prefix %s started", prefix)
+	for _, id := range triggers {
+		err := database.RemoveTrigger(id)
+		if err != nil {
+			return err
```
kissken, line 25: maybe better wrap err?

kissken (commented):
left some comments

database/redis/trigger_test.go
kissken: we already it test file

cmd/cli/triggers.go
Dimedrolity: yes, noted in [comment](https://github.com/moira-alert/moira/pull/790/commits/ecd22470e6d393f0cfa7b53fbee551a53c199231#diff-f463a96da8e66f155de89be8fc4d0ecde2eb0f6e556a4cc5642b0d3c0e5df5deR153)

cmd/cli/triggers.go
Dimedrolity: yes, I will do, because there are 2 `return err` cases

cmd/cli/triggers.go
kissken: think that can be misunderstanding, lets think about naming, cause notion prefix in Moira already use, ya?

database/redis/trigger_test.go
Dimedrolity: please read the [commit message](https://github.com/moira-alert/moira/pull/790/commits/b3d00e6d69a800be314e00f8b1423e1b4e25a988)

kissken: ah, missed this line, ok

cmd/cli/triggers.go
Dimedrolity: Why? This const is only for one function, so it is local const

cmd/cli/triggers.go
Dimedrolity: from the [docs](https://redis.io/docs/manual/pipelining/): pipeline is used for improving performance. So pipe for this case is premature optimization, command works pretty fast

cmd/cli/triggers.go
Dimedrolity: added both, [first ](https://github.com/moira-alert/moira/pull/790/commits/de53e6d0e9653797804aaec24236aa1765c25897#diff-bbe74c5f7c7c4e46d51dec5dfa5561be8b5fd6f26ae4ea6eac8bc27876f7d98bR13) and [second](https://github.com/moira-alert/moira/pull/790/commits/de53e6d0e9653797804aaec24236aa1765c25897#diff-bbe74c5f7c7c4e46d51dec5dfa5561be8b5fd6f26ae4ea6eac8bc27876f7d98bR26)

cmd/cli/triggers.go
Dimedrolity: do you mean prefix tree?

kissken: yep, as least

Dimedrolity: I don't agree about misunderstanding, prefix is common term for programming. So prefix means some symbols at start of string. There is no reference to tree in method name or comment of GetTriggerIDsByPrefix

Dimedrolity: renamed for GetTriggerIDsStartWith(prefix string)

kissken (approved):
Let’s make release, but rolling with Mikhail (he has rights to do this)
and with Alexey
and maybe ask to do backup before run

PR #793, opened 2022-10-26
net/http library's consts for representing http codes

maksgalimz:
# Http codes

Using http package consts instead bare numbers for http codes

### DIFF
--- a/api/controller/contact_test.go
+++ b/api/controller/contact_test.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"errors"
 	"fmt"
+	"net/http"
 	"strings"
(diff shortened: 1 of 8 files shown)


kissken:
will other pkg be in another PR?

maksgalimz:
what other pkg?

kissken:
example:
https://github.com/moira-alert/moira/blob/master/metric_source/remote/request.go#L45
https://github.com/moira-alert/moira/blob/master/senders/victorops/api/alert.go#L60

kissken (approved):
uhuuuu!

PR #800, opened 2022-11-08
Update carbonapi version

Tetrergeru:
Updated carbonapi version due to panics caused by older on some invalid targets.

Added error unwrapping, because errors are wrapped in newer versions of carbonapi. 

### Can not be merged

Due to another bug in carbonapi that makes checker panic on existing valid triggers.


test_res.txt
```diff
@@ -0,0 +1,528 @@
+ok  	github.com/moira-alert/moira	(cached)
```
kissken, line 1: hmmm, added into gitignore?

Tetrergeru: I acidently added this file to git. Deleted :ok_hand:

Dimedrolity:
>Updated carbonapi version due to panics caused by older on some invalid targets.

@Tetrergeru What the reason is mean? Please explain where the panics were on old version of carbonapi

metric_source/local/errors.go
```diff
@@ -33,7 +34,7 @@ func (err ErrUnknownFunction) Error() string {
 
 // isErrUnknownFunction checks error for carbonapi.errUnknownFunction
 func isErrUnknownFunction(err error) bool {
-	switch err.(type) {
+	switch merry.Unwrap(err).(type) {
```
kissken, line 37: missed any tests on this case?

let's add?

Tetrergeru: This change was caused by package update. Without it trigger tests start to fail

Tetrergeru:
@Dimedrolity, old version of carbonapi successfully parses invalid target and then panics on Metrics method call

Dimedrolity:
@Tetrergeru Ok, thanks. Provide example of invalid target, please

Tetrergeru:
@Dimedrolity, the bug was discovered with target `aliasByNode(movingMin(Infrastructure.ClusterConfig.cloud.ClusterConfig_SloMeter.nonEquivalent.V2.*), '10min'), 6)`

Dimedrolity:
@Tetrergeru Ok, I suggest to add test to that target to ensure that bug was fixed. It seems need to test _metric_source/local/local.go_ Fetch method, because only there is the call to Metrics method

kissken:
@Dimedrolity panic is not primary reason for update carbon-api, it is secondary and background. And PR have necessary title :)

Dimedrolity:
@kissken I don't see another reason in title or description of PR. So what is the primary reason for update carbon-api?

kissken:
@Dimedrolity technical debt

Dimedrolity:
@kissken Ok, thanks.

Do you think [this is overhead](https://github.com/moira-alert/moira/pull/800#issuecomment-1309767672)?

kissken:
@Dimedrolity aha :)

Dimedrolity (approved):
Well

kissken:
buuuut @Tetrergeru let's mark as draft, end explain, why we cannot merge, please

Tetrergeru:
@kissken, Done

Dimedrolity:
> Due to another bug in carbonapi that makes checker panic on existing valid triggers.

Oh😒
How many existing valid triggers are panic? Why they panic?

kissken:
@Tetrergeru heeeey, let's close this PR? %)

PR #801, opened 2022-11-09
Add trigger ID validation

Tetrergeru:
User could crate triggers with custom name that contains invalid characters. Added validation in trigger controller.

Add tests for cases with correct and incorrect trigger ID.

### DIFF
--- a/api/controller/triggers.go
+++ b/api/controller/triggers.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"fmt"
 	"math"
+	"regexp"
 
 	"github.com/gofrs/uuid"
 
@@ -13,6 +14,8 @@ import (
 )
 
 const pageSizeUnlimited int64 = -1
+
+var idValidationPattern = regexp.MustCompile(`^[a-zA-Z0-9\-_]+$`)
 
 // CreateTrigger creates new trigger
 func CreateTrigger(dataBase moira.Database, trigger *dto.TriggerModel, timeSeriesNames map[string]bool) (*dto.SaveTriggerResponse, *api.ErrorResponse) {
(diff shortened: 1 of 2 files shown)


kissken (approved):

api/controller/triggers.go
```diff
@@ -14,6 +15,8 @@ import (
 
 const pageSizeUnlimited int64 = -1
 
+var idValidationPattern = regexp.MustCompile(`^[a-zA-Z0-9\-_]+$`)
```
Dimedrolity, line 18: It is just question without suggestion: why you declared `idValidationPattern` at package level scope and not in `CreateTrigger` func? There is only one usage of var

api/controller/triggers_test.go
```diff
@@ -47,6 +47,31 @@ func TestCreateTrigger(t *testing.T) {
...
+		triggerModel := dto.TriggerModel{ID: triggerID}
+		dataBase.EXPECT().GetTrigger(triggerModel.ID).Return(moira.Trigger{}, database.ErrNil)
+		dataBase.EXPECT().AcquireTriggerCheckLock(gomock.Any(), 10)
+		dataBase.EXPECT().DeleteTriggerCheckLock(gomock.Any())
+		dataBase.EXPECT().GetTriggerLastCheck(gomock.Any()).Return(moira.CheckData{}, database.ErrNil)
+		dataBase.EXPECT().SetTriggerLastCheck(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil)
+		dataBase.EXPECT().SaveTrigger(gomock.Any(), triggerModel.ToMoiraTrigger()).Return(nil)
+		resp, err := CreateTrigger(dataBase, &triggerModel, make(map[string]bool))
+		So(err, ShouldBeNil)
+		So(resp.Message, ShouldResemble, "trigger created")
```
Dimedrolity, line 61: JFYI `ShouldEqual` is enough, but it's ok, nevermind

api/controller/triggers.go
Tetrergeru: Because regex compilation is a heavy-weight operation. Our controller functions don't have any state, so I had to extract regex to global variable

api/controller/triggers_test.go
Tetrergeru: I used already existing tests as a template. It used `ShouldResemble` in the same case

api/controller/triggers.go
Dimedrolity: Nice, thank you!

api/controller/triggers_test.go
Dimedrolity: Ok, I did so also

Dimedrolity (approved):

PR #802, opened 2022-11-10
n checkers

maksgalimz:
Ticket brings an ability to launch more than one checker

### DIFF
--- a/checker/config.go
+++ b/checker/config.go
@@ -16,4 +16,6 @@ type Config struct {
 	LogFile                     string
 	LogLevel                    string
 	LogTriggersToLevel          map[string]string
+	MetricEventPopBatchSize     int
+	MetricEventPopDelay         time.Duration
 }
--- a/checker/config.go
+++ b/checker/config.go
@@ -16,6 +16,6 @@ type Config struct {
 	LogFile                     string
 	LogLevel                    string
 	LogTriggersToLevel          map[string]string
-	MetricEventPopBatchSize     int
+	MetricEventPopBatchSize     int64
 	MetricEventPopDelay         time.Duration
 }
--- a/checker/worker/metrics.go
+++ b/checker/worker/metrics.go
@@ -61,7 +61,7 @@ func (worker *Checker) addRemoteTriggerIDsIfNeeded(triggerIDs []string) {
 
 func (worker *Checker) getTriggerIDsToCheck(triggerIDs []string) []string {
 	lazyTriggerIDs := worker.lazyTriggerIDs.Load().(map[string]bool)
-	var triggerIDsToCheck []string
+	var triggerIDsToCheck []string = make([]string, 0, len(triggerIDs))
 	for _, triggerID := range triggerIDs {
 		if _, ok := lazyTriggerIDs[triggerID]; ok {
 			randomDuration := worker.getRandomLazyCacheDuration()
--- a/checker/worker/metrics.go
+++ b/checker/worker/metrics.go
@@ -61,7 +61,7 @@ func (worker *Checker) addRemoteTriggerIDsIfNeeded(triggerIDs []string) {
 
 func (worker *Checker) getTriggerIDsToCheck(triggerIDs []string) []string {
 	lazyTriggerIDs := worker.lazyTriggerIDs.Load().(map[string]bool)
-	triggerIDsToCheck := make([]string, len(triggerIDs))
+	var triggerIDsToCheck []string
 	for _, triggerID := range triggerIDs {
 		if _, ok := lazyTriggerIDs[triggerID]; ok {
 			randomDuration := worker.getRandomLazyCacheDuration()
--- a/checker/worker/worker.go
+++ b/checker/worker/worker.go
@@ -50,7 +50,7 @@ func (worker *Checker) Start() error {
 	}
 
 	subscribeMetricEventsParams := moira.SubscribeMetricEventsParams{
-		BatchSize: int64(worker.Config.MetricEventPopBatchSize),
+		BatchSize: worker.Config.MetricEventPopBatchSize,
 		Delay:     worker.Config.MetricEventPopDelay,
 	}
 
--- a/checker/worker/worker.go
+++ b/checker/worker/worker.go
@@ -43,7 +43,18 @@ func (worker *Checker) Start() error {
 
 	worker.lastData = time.Now().UTC().Unix()
 
-	metricEventsChannel, err := worker.Database.SubscribeMetricEvents(&worker.tomb)
+	if worker.Config.MetricEventPopBatchSize == 0 {
+		worker.Config.MetricEventPopBatchSize = 100
+	} else if worker.Config.MetricEventPopBatchSize < 0 {
+		return errors.New("MetricEventPopBatchSize param less than zero")
+	}
+
+	subscribeMetricEventsParams := moira.SubscribeMetricEventsParams{
+		BatchSize: int64(worker.Config.MetricEventPopBatchSize),
+		Delay:     worker.Config.MetricEventPopDelay,
+	}
+
+	metricEventsChannel, err := worker.Database.SubscribeMetricEvents(&worker.tomb, &subscribeMetricEventsParams)
 	if err != nil {
 		return err
 	}
--- a/database/redis/database.go
+++ b/database/redis/database.go
@@ -25,6 +25,7 @@ const (
 
 const (
 	receiveErrorSleepDuration = time.Second
+	receiveEmptySleepDuration = time.Second
 )
 
 // DBSource is type for describing who create database instance
--- a/database/redis/database.go
+++ b/database/redis/database.go
@@ -2,8 +2,6 @@ package redis
 
 import (
 	"context"
-	"fmt"
-	"io"
 	"time"
 
 	"github.com/moira-alert/moira/clock"
@@ -13,19 +11,13 @@ import (
 	"github.com/go-redsync/redsync/v4/redis/goredis/v8"
 	"github.com/moira-alert/moira"
 	"github.com/patrickmn/go-cache"
-	"gopkg.in/tomb.v2"
 )
 
-const pubSubWorkerChannelSize = 16384
+const metricEventChannelSize = 16384
 
 const (
 	cacheCleanupInterval         = time.Minute * 60
 	cacheValueExpirationDuration = time.Minute
-)
-
-const (
-	receiveErrorSleepDuration = time.Second
-	receiveEmptySleepDuration = time.Second
 )
 
 // DBSource is type for describing who create database instance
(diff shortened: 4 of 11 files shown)


checker/worker/metrics.go
```diff
@@ -61,7 +61,7 @@ func (worker *Checker) addRemoteTriggerIDsIfNeeded(triggerIDs []string) {
 
 func (worker *Checker) getTriggerIDsToCheck(triggerIDs []string) []string {
 	lazyTriggerIDs := worker.lazyTriggerIDs.Load().(map[string]bool)
-	triggerIDsToCheck := make([]string, len(triggerIDs))
+	var triggerIDsToCheck []string
```
kissken, line 64: hm, why?

maksgalimz: Each empty cell produce trigger processing, that of course will be prevented on some step, but this produces log records those make a mess in logs

database/redis/metric.go
```diff
@@ -106,6 +106,7 @@ func (connector *DbConnector) SaveMetrics(metrics map[string]*moira.MatchedMetri
 	ctx := connector.context
 
 	rand.Seed(time.Now().UnixNano())
+	pipe := c.TxPipeline()
```
kissken, line 109: and changed we filters together?

database/redis/metric.go
```diff
@@ -136,40 +137,92 @@ func (connector *DbConnector) SaveMetrics(metrics map[string]*moira.MatchedMetri
...
+		}()
+	}
+
+	return metricChannel, nil
+}
+
+func (connector *DbConnector) handlePopResponse(data []string, popError error, responseChannel chan string) time.Duration {
+	if popError != nil {
+		if popError != redis.Nil {
+			connector.logger.Errorf("Error happened: %s.", popError)
```
kissken, line 214: maybe more give more info?

checker/worker/metrics.go
Dimedrolity: It is strange that there was instantiating of slice with len of triggers and then appending IDs to it, as a result we have slice with N empty strings and M triggers IDs to check (M <= N). @maksgalimz Is it was just a bug?

database/redis/metric_test.go
```diff
@@ -443,6 +443,7 @@ func TestMetricSubscription(t *testing.T) {
 		So(err, ShouldBeNil)
 		err = dataBase.SaveMetrics(map[string]*moira.MatchedMetric{metric2: met2})
 		So(err, ShouldBeNil)
+		<-time.After(time.Second * 6)
```
Dimedrolity, line 446:
```suggestion
		time.Sleep(time.Second * 6)
```

Dimedrolity: I suggest to avoid channels if there is no need

Dimedrolity: And question: Why this test passes only if use 6 sec delay? I try to run without delay and it fails, but why? And why it not fails before "n checkers" feature?

database/redis/metric.go
```diff
@@ -136,40 +137,92 @@ func (connector *DbConnector) SaveMetrics(metrics map[string]*moira.MatchedMetri
...
+	ctx := connector.context
+	c := *connector.client
+
+	if err := c.Ping(ctx).Err(); err != nil {
 		return nil, err
 	}
 
+	go func() {
+		<-tomb.Dying()
+		close(responesChannel)
```
kissken, line 171: maybe better use defer construct?

checker/worker/metrics.go
Dimedrolity:
```suggestion
	triggerIDsToCheck := make([]string, 0, len(triggerIDs))
```

Dimedrolity: Maybe author wanted to set capacity of slice

database/redis/metric.go
```diff
@@ -136,40 +137,92 @@ func (connector *DbConnector) SaveMetrics(metrics map[string]*moira.MatchedMetri
...
 				continue
 			}
-			metricsChannel <- metricEvent
+			metricChannel <- metricEvent
 		}
 	}()
 
-	return metricsChannel, nil
+	for channelIdx := 0; channelIdx < len(metricEventsChannels); channelIdx++ {
+		metricEventsChannel := metricEventsChannels[channelIdx]
```
kissken, line 192: but this is not channel as channel... 
i saw that this was naming before urs changes, but hm...

not necessary, i think

kissken (changes requested):
left little comments

checker/worker/metrics.go
maksgalimz: Agree

database/redis/metric.go
maksgalimz: Didn't get the question

database/redis/metric_test.go
maksgalimz: Before this feature polling was based on PubSub subscription and kill() operation within Tomb sent unsubscribe message to redis. That model gave time while unsubscribe message processing to finish other queue messages. Now killing happens immediately

database/redis/metric.go
maksgalimz: defer construction will call this function in the same gorutine, that will block program execution

database/redis/metric.go
kissken: this changes will be affected filter service or not?

database/redis/metric.go
maksgalimz: Redis documentation name it channel, so i think it is correct

database/redis/metric.go
maksgalimz: okey

database/redis/metric.go
kissken: it's may me my misunderstanding...

maksgalimz: Yes, changes will affect filter and checker services

database/redis/metric.go
```diff
@@ -136,40 +137,92 @@ func (connector *DbConnector) SaveMetrics(metrics map[string]*moira.MatchedMetri
...
+		connector.logger.Errorf("Sending metric event error. Error: %v;", err)
+		return err
+	}
+
 	return nil
 }
 
+const (
+	metricEventPopBatchSize   = 100
+	receiveEmptySleepDuration = time.Second
```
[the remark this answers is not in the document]
maksgalimz: I think receiveEmptySleepDuration param is not important for configuration. BatchSize can be important. So I extracted this param, also I added new param for this method.

Tetrergeru (dismissed):

checker/config.go
```diff
@@ -16,4 +16,6 @@ type Config struct {
 	LogFile                     string
 	LogLevel                    string
 	LogTriggersToLevel          map[string]string
+	MetricEventPopBatchSize     int
```
kissken, line 19: let's add also in [config](https://github.com/moira-alert/moira/blob/master/pkg/notifier/notifier.yml), these two params please?)

checker/worker/worker.go
```diff
@@ -43,7 +43,18 @@ func (worker *Checker) Start() error {
...
 
-	metricEventsChannel, err := worker.Database.SubscribeMetricEvents(&worker.tomb)
+	if worker.Config.MetricEventPopBatchSize == 0 {
+		worker.Config.MetricEventPopBatchSize = 100
+	} else if worker.Config.MetricEventPopBatchSize < 0 {
+		return errors.New("MetricEventPopBatchSize param less than zero")
+	}
+
+	subscribeMetricEventsParams := moira.SubscribeMetricEventsParams{
+		BatchSize: int64(worker.Config.MetricEventPopBatchSize),
```
kissken, line 53: maybe [here](https://github.com/moira-alert/moira/pull/802/files#diff-a1c3b1ad9fc3f8e561060748f53a8cee5af1448c3f02bcf07537263d26d256e8R19) int64?

kissken: to not cast in each loop)

maksgalimz: There is no loop

checker/config.go
maksgalimz: Okey

checker/worker/worker.go
kissken: aaaaa, sooorry(( my fault

kissken (approved):

database/redis/metric.go
```diff
@@ -136,40 +137,87 @@ func (connector *DbConnector) SaveMetrics(metrics map[string]*moira.MatchedMetri
...
 		}
 	}()
 
-	return metricsChannel, nil
+	for channelIdx := 0; channelIdx < len(metricEventsChannels); channelIdx++ {
+		metricEventsChannel := metricEventsChannels[channelIdx]
+		go func() {
+			var popDelay time.Duration
+			for {
+				startPop := time.After(popDelay)
```
Dimedrolity, line 191: Would you please explain, why we need delay and batchSize? Is it performance optimization?

database/redis/metric.go
```diff
@@ -136,40 +137,87 @@ func (connector *DbConnector) SaveMetrics(metrics map[string]*moira.MatchedMetri
...
 }
 
 // SubscribeMetricEvents creates subscription for new metrics and return channel for this events
-func (connector *DbConnector) SubscribeMetricEvents(tomb *tomb.Tomb) (<-chan *moira.MetricEvent, error) {
-	metricsChannel := make(chan *moira.MetricEvent, pubSubWorkerChannelSize)
-	dataChannel, err := connector.manageSubscriptions(tomb, metricEventsChannels)
-	if err != nil {
+func (connector *DbConnector) SubscribeMetricEvents(tomb *tomb.Tomb, params *moira.SubscribeMetricEventsParams) (<-chan *moira.MetricEvent, error) {
+	responesChannel := make(chan string, pubSubWorkerChannelSize)
+	metricChannel := make(chan *moira.MetricEvent, pubSubWorkerChannelSize)
```
Dimedrolity, line 155: How do you think, maybe rename `pubSubWorkerChannelSize` and `pubSubWorkerChannelSize`, because we don't use pub/sub model anymore? Or now we continue to use pub/sub model but with Queue as implementation?

database/redis/metric.go
Dimedrolity, line 109: Would you please explain, why you use pipe here? Is it perfomance optimization?

database/redis/metric.go
```diff
@@ -136,40 +137,87 @@ func (connector *DbConnector) SaveMetrics(metrics map[string]*moira.MatchedMetri
...
+
+	return metricChannel, nil
+}
+
+func (connector *DbConnector) handlePopResponse(data []string, popError error, responseChannel chan string, defaultDelay time.Duration) time.Duration {
+	if popError != nil {
+		if popError != redis.Nil {
+			connector.logger.Errorf("Failed to pop new metric events. Error: %s.", popError)
+		}
+		return receiveErrorSleepDuration
```
Dimedrolity, line 211: Would use please explain, why we need to have specific delay on error and specific delay on empty? I think default delay is enough, but I can mistake

Dimedrolity (commented):
Well. My review is ready to check, please check it

checker/worker/worker.go
```diff
@@ -43,7 +43,18 @@ func (worker *Checker) Start() error {
 
 	worker.lastData = time.Now().UTC().Unix()
 
-	metricEventsChannel, err := worker.Database.SubscribeMetricEvents(&worker.tomb)
+	if worker.Config.MetricEventPopBatchSize == 0 {
```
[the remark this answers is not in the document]
maksgalimz: I think that getSettings method is better place for processing program defaults, but moving them will break current pattern program follows (Above same logic for MaxParallelChecks param). I don't want to break pattern or especially change pattern.

database/redis/metric_test.go
maksgalimz: You are right

database/redis/metric.go
```diff
@@ -136,40 +137,87 @@ func (connector *DbConnector) SaveMetrics(metrics map[string]*moira.MatchedMetri
...
 	return nil
 }
 
 // SubscribeMetricEvents creates subscription for new metrics and return channel for this events
-func (connector *DbConnector) SubscribeMetricEvents(tomb *tomb.Tomb) (<-chan *moira.MetricEvent, error) {
-	metricsChannel := make(chan *moira.MetricEvent, pubSubWorkerChannelSize)
-	dataChannel, err := connector.manageSubscriptions(tomb, metricEventsChannels)
-	if err != nil {
+func (connector *DbConnector) SubscribeMetricEvents(tomb *tomb.Tomb, params *moira.SubscribeMetricEventsParams) (<-chan *moira.MetricEvent, error) {
+	responesChannel := make(chan string, pubSubWorkerChannelSize)
```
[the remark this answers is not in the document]
maksgalimz: ok

database/redis/metric.go
maksgalimz: One request that contains 100 records is better than 100 singular request. Delay helps us to control amount of request to redis per time slot

database/redis/metric.go
maksgalimz: Agree

database/redis/metric.go
maksgalimz: Yes it's a performance optimization. Pipe collects records and send them together to redis. That model minimizes internet communicate overheads.

database/redis/metric.go
maksgalimz: The separation can help to adjust program better.

database/redis/database.go
```diff
@@ -25,6 +25,7 @@ const (
 
 const (
 	receiveErrorSleepDuration = time.Second
```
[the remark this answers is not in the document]
maksgalimz: ok

database/redis/metric.go
Dimedrolity: I don't like pipe because people use it everywhere even it is not need like premature optimization. And in Moira pipe is used everywhere, it is mixed up with business logic pretty often. So it makes fragile architecture, mixed data access layer and business logic layer. But it is IMHO, nevermind, just my thoughs

database/redis/metric.go
Dimedrolity: I don't like premature optimization, but it is fine. Generally I think if you make optimization in advance, then you will not know if this solution was realy need

database/redis/metric.go
Dimedrolity: Yes, I understand it. But the question was about problem, not about advantage of solution

Dimedrolity (approved):

kissken (approved):

PR #803, opened 2022-11-10
n checkers v2

maksgalimz:
one metric events handler for all checker


kissken:
@maksgalimz hello, now should we close it?

kissken:
see https://github.com/moira-alert/moira/pull/802

Issue #797, opened 2022-11-07
Refactor Senders configuration


Dimedrolity:
## Overview

We have Sender interface with Init method that configures Sender.

```
// Sender interface for implementing specified contact type sender
type Sender interface {
	SendEvents(events NotificationEvents, contact ContactData, trigger TriggerData, plot [][]byte, throttled bool) error
	Init(senderSettings map[string]string, logger Logger, location *time.Location, dateTimeFormat string) error
}
```

## Problem

It is strange approach to configure entity, each Sender can have its own settings and dependencies. 

Now we have `logger, location *time.Location, dateTimeFormat string`, I think authors (developers) wanted to don't repeat yourself and created that unified Init method. Maybe will be better extract BaseConfig struct and maybe Config struct for each Sender.

If we follow [DI principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle), then constructor func (that have dependencies) and Init method in one time is not well, both methods goal is to configure. 

## Solution

Need to create constructor functions for each Sender, and move init logic from Init method to constructors, then remove Init from interface and impelementrations.

PR #804, opened 2022-11-11
Refactor Senders initialization

Dimedrolity:
Closes https://github.com/moira-alert/moira/issues/797


Dimedrolity:
Want to move senders directory to notifier directory, because senders have no sense without notifier. Also errors.go in the root of Moira contains Notifier erros, should be also moved to notifier dir.

Dimedrolity:
- [ ] Edit [documentation](https://moira.readthedocs.io/en/latest/development/notifier.html?highlight=sender#writing-your-own-notification-sender:~:text=with%20appropriate%20type.-,Sender,settings%20in%20common%20config%2C%20which%20is%20passed%20to%20the%20Init%20method.,-Next) after merge

notifier/senders/msteams/msteams.go
```diff
@@ -44,24 +45,27 @@ type Sender struct {
...
 	sender.location = location
 	sender.frontURI = senderSettings["front_uri"]
 	maxEvents, err := strconv.Atoi(senderSettings["max_events"])
 	if err != nil {
-		return fmt.Errorf("max_events should be an integer: %w", err)
+		return nil, fmt.Errorf("max_events should be an integer: %w", err)
 	}
 	sender.maxEvents = maxEvents
 	sender.client = &http.Client{
 		Timeout: time.Duration(30) * time.Second, //nolint
```
kissken, line 61: let's move it into const

notifier/senders/script/script.go
```diff
@@ -7,41 +7,44 @@ import (
...
-type scriptNotification struct {
-	Events    []moira.NotificationEvent `json:"events"`
-	Trigger   moira.TriggerData         `json:"trigger"`
-	Contact   moira.ContactData         `json:"contact"`
-	Throttled bool                      `json:"throttled"`
-	Timestamp int64                     `json:"timestamp"`
-}
+// NewSender creates Sender instance.
+func NewSender(senderSettings map[string]string, logger moira.Logger) (*Sender, error) {
+	sender := &Sender{}
```
kissken, line 23: can be ini after validate

notifier/senders/selfstate/selfstate.go
```diff
@@ -2,25 +2,30 @@ package selfstate
...
 type Sender struct {
 	Database moira.Database
 	logger   moira.Logger
 }
 
-// Init read yaml config
-func (sender *Sender) Init(senderSettings map[string]string, logger moira.Logger, location *time.Location, dateTimeFormat string) error {
+// NewSender creates Sender instance.
+func NewSender(logger moira.Logger, db moira.Database) *Sender {
+	sender := &Sender{
```
kissken, line 18: simplify be

return &Sender{Database: db, logger logger}

notifier/senders/webhook/webhook_test.go
```diff
@@ -33,16 +31,19 @@ func TestSender_SendEvents(t *testing.T) {
...
 					if err != nil {
 						w.WriteHeader(status)
-						w.Write([]byte(err.Error())) //nolint
+						_, err = w.Write([]byte(err.Error()))
+						So(err, ShouldBeNil)
 					}
 					status, err = testRequestBody(r)
 					if err != nil {
-						w.Write([]byte(err.Error())) //nolint
+						_, err = w.Write([]byte(err.Error())) //nolint
```
kissken, line 45: seems //nolint dont need more

kissken (changes requested):
left some comments

notifier/senders/mail/mail.go
```diff
@@ -30,18 +31,24 @@ type Sender struct {
 	dateTimeFormat string
 }
 
-// Init read yaml config
-func (sender *Sender) Init(senderSettings map[string]string, logger moira.Logger, location *time.Location, dateTimeFormat string) error {
+// NewSender creates Sender instance.
```
Tetrergeru, line 34: New<Something> seems th be the conventional way to name constructor in golang, thus these commants are redundant

notifier/senders/mail/mail.go
```diff
@@ -30,18 +31,24 @@ type Sender struct {
...
+func NewSender(senderSettings map[string]string, logger moira.Logger, location *time.Location, dateTimeFormat string) (*Sender, error) {
```
Tetrergeru, line 35: Maybe NewSender methods should return interfaces instance instead of structure instance, so that we could hide the implementation inside the package

Tetrergeru (changes requested):

notifier/senders/mail/mail.go
Dimedrolity: It is interesting, I like your idea, so caller will have access only to interface methods and can't access to other methods and struct fields, but it breaks some tests (at least mattermost/internal_test.go). So I think if do that refactor, then need to refactor all Senders, but now it breaks, so it is not one step refactor and should be done in another PR. And I found that returning interface is controversial, discussion https://stackoverflow.com/questions/59040392/how-to-write-idiomatic-constructor and rule "accepting interfaces and returning structs" https://tutorialedge.net/golang/accept-interfaces-return-structs/

Tetrergeru: I believe in this case it woukld be acceptable to return interface because:
1) It is not a library with a public api
2) The only thing we do with the return value is that we assign it to a variable with an interface type

On the other hand, there is no point in hiding implementation precisely because it is not a library with a public api :-)

So I am resolving the issue

Tetrergeru (approved):

notifier/senders/mail/mail.go
Dimedrolity: Lets wait for answer from Xenia, maybe we will remove it

notifier/registrator.go
```diff
@@ -40,52 +40,60 @@ const (
 	mattermostSender  = "mattermost"
 )
 
-// RegisterSenders watch on senders config and register all configured senders
+// RegisterSenders creates all senders and registers them.
 func (notifier *StandardNotifier) RegisterSenders(connector moira.Database) error { //nolint
```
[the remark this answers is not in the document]
kissken: and can return error...

integration_tests/notifier/notifier_test.go
```diff
@@ -78,41 +79,51 @@ var event = moira.NotificationEvent{
...
+		err := database.SaveContact(&contact)
+		So(err, ShouldBeNil)
+		err = database.SaveSubscription(&subscription)
+		So(err, ShouldBeNil)
+		err = database.SaveTrigger(trigger.ID, &trigger)
+		So(err, ShouldBeNil)
+		err = database.PushNotificationEvent(&event, true)
+		So(err, ShouldBeNil)
+
+		notifier2 := notifier.NewNotifier(database, logger, notifierConfig, notifierMetrics, metricsSourceProvider, map[string]moira.ImageStore{})
```
kissken, line 95: let's make pretty name, please

notifier/senders/discord/sender.go
```diff
@@ -26,25 +27,28 @@ type Sender struct {
 	botUserID string
 }
 
-// Init reads the yaml config
-func (sender *Sender) Init(senderSettings map[string]string, logger moira.Logger, location *time.Location, dateTimeFormat string) error {
-	var err error
+// NewSender creates Sender instance.
+func NewSender(senderSettings map[string]string, logger moira.Logger, location *time.Location, db moira.Database) (*Sender, error) {
```
kissken, line 31: has it any tests?

kissken:
**merge after release**

notifier/registrator.go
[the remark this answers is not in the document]
Dimedrolity: I think testing this func is overengineering. If we test this func, then we increment some coverage results, but what is the profit? Do we increase reliablitiy? I think no, because every senders consctructor is tested in place. To test `RegisterSenders` func, we have to cover N cases (for each Sender), but it is repeating of constructors tests, so it is redundant

integration_tests/notifier/notifier_test.go
Dimedrolity: Done https://github.com/moira-alert/moira/pull/804/commits/c6e66a55d3e503cb001cca25379dee5f6a46182e

notifier/senders/discord/sender.go
Dimedrolity: Yes, there is some tests

notifier/senders/msteams/msteams.go
kissken: not in func, please :)

Dimedrolity: I just drop `//nolint`, it is not necessary to refactor all Moira code. This const `timeout` is so little thing compare to problems that we have in Moira, so I advise you don't pay attention to it.

Dimedrolity:
We don't need `pretty`, we need `self descriptive`. These messages are self descriptive enough. It will be squashed in one commit on merge anyway

Tetrergeru (approved):

PR #806, opened 2022-11-18
fix(api): 500 on incorrect requests

Tetrergeru:
Api used to return 500 on create/update/check trigger requests with incorrect payload. Now it returns 400.

I also chaked all other handlers to make shure they return 400 in this case.


kissken (approved):
👍

Dimedrolity (approved):

kissken:
@Tetrergeru let's up this pull, please? c:

PR #808, opened 2022-12-01
Fix target validation

Dimedrolity:
There was a bug: user could create trigger with target with space in metric name like `a b`. Now validation doesn't allow it.


Dimedrolity:
Result is 
image

kissken (approved):

Tetrergeru (approved):

PR #811, opened 2022-12-06
Add logger features for convenient creation of structured logs

Tetrergeru:
# PR Summary

Additional information

Closes/Relates #issue


interfaces.go
```diff
@@ -168,15 +168,29 @@ type Mutex interface {
 // Logger implements logger abstraction
 type Logger interface {
 	Debug(args ...interface{})
+	DebugWithError(msg string, err error)
 	Debugf(format string, args ...interface{})
+	Debugb() EventBuilder
```
Dimedrolity, line 173: Why new methods with `b` return `EventBuilder`, but methods `WithError` are not?

interfaces.go
```diff
@@ -190,6 +204,16 @@ type Logger interface {
 	Clone() Logger
 }
 
+type EventBuilder interface {
```
Dimedrolity, line 207: Lets add godoc comment what is this, maybe with mention what type this interface covers like here https://github.com/moira-alert/moira/blob/master/senders/mattermost/client.go#L5

logging/zerolog_adapter/event_builder.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/moira-alert/moira"
+	"github.com/rs/zerolog"
+)
+
+type EventBuilder struct {
+	*zerolog.Event
+}
+
+func (e EventBuilder) Msg(msg string) {
+	if e.Event != nil {
```
Dimedrolity, line 15: Do we really need to check for Event in every method? Maybe it can be moved to constructor

interfaces.go
Dimedrolity: Maybe move interfaces `EventBuilder` and `Logger` to logging pkg. Or move at least `EventBuilder` . So it will be more near to usage. I don't like interfaces.go in root of project, it is trash I consider. Is it fine, how do you think?

Dimedrolity (commented):
Well

kissken:
Should we mark old methods as deprecated? 
And make one more PR to drop old methods and use only new?

interfaces.go
Tetrergeru: Because `WithError` is a convenient shorthand to write logs with error messages in one line.

Also `Msg` should be called when all the tags are set, and `WithError` calls `Msg` , thus no more tags can be set

Tetrergeru:
> Should we mark old methods as deprecated? And make one more PR to drop old methods and use only new?

I think that we can deprecate (and later delete) `-f` methods because we don't have use for them now.

As for the plain `Debug`, `Info` e.t.c., I propose, we change their signature from `Debug(args ...interface{})` to `Debug(msg string)`. It would we useful because sometimes we need a plai text log without any tags

kissken:
@Tetrergeru may be do it in this PR (mark)?

logging/zerolog_adapter/event_builder.go
```diff
@@ -0,0 +1,57 @@
+package logging
```
kissken, line 1: let's add godoc

logging/zerolog_adapter/logger.go
```diff
@@ -79,6 +79,12 @@ func (l Logger) Debug(args ...interface{}) {
 	event.Timestamp().Msg(fmt.Sprint(args...))
 }
 
+func (l Logger) DebugWithError(msg string, err error) {
```
kissken, line 82: why we cannot use builder for this situation?

kissken: i mean without extra func

Tetrergeru: We can, but it is quite a common case, so I made additional methods for cleaner code

kissken: we will move on builder in end, won't it? 

or we save two logger components event + simple log, after we drop func is nor use?

interfaces.go
Tetrergeru: Added godoc.

logging/zerolog_adapter/event_builder.go
```diff
@@ -0,0 +1,57 @@
...
+func (e EventBuilder) String(key, value string) logging.EventBuilder {
+	if e.Event != nil {
+		e.Event.Str(key, value)
+	}
+	return e
+}
+
+func (e EventBuilder) Error(err error) logging.EventBuilder {
+	if e.Event != nil {
+		e.Event.Str("error", err.Error())
```
kissken, line 29: may be error use instead of Str

logging/zerolog_adapter/event_builder.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func (e EventBuilder) Int64(key string, value int64) logging.EventBuilder {
+	if e.Event != nil {
+		e.Event.Int64(key, value)
+	}
+	return e
+}
+
+func (e EventBuilder) Value(key string, value interface{}) logging.EventBuilder {
```
kissken, line 48: may be rename into Interface

logging/zerolog_adapter/event_builder.go
```diff
@@ -0,0 +1,57 @@
...
+	return e.String(key, fmt.Sprintf("%v", value))
```
kissken, line 49: and may be use Interface

logging/zerolog_adapter/event_builder.go
Dimedrolity: We don't want panic I think. And what if we add constructor with error in return parameter? Is it will better or maybe it is overhead? These error checks in every method are strange repeats I think, but I can be wrong, maybe it is fine for Moira or for Golang

interfaces.go
Tetrergeru: `tag` is a common name for all key-value pairs set for a sertain event.

`WithError` methods were removed after duscussion with @kissken.

Added comment to Msg in EventBuilder interface.

interfaces.go
Tetrergeru: Created task

logging/zerolog_adapter/event_builder.go
Tetrergeru: I looked to the source code of zerolog library and discovered that `event == nil` only in case of disabled logs. So we shouldn't treat this as an error and rather just filter `nil` case in all consecutive method calls

logging/zerolog_adapter/logger.go
Tetrergeru: Removed these method after discussion

logging/zerolog_adapter/event_builder.go
Tetrergeru: Changed

logging/zerolog_adapter/event_builder.go
Tetrergeru: Renamed method `Value` to `Interface`

logging/zerolog_adapter/event_builder.go
Tetrergeru: Done!

Tetrergeru:
Old method for logging were deleted @kissken

Dimedrolity:
@Tetrergeru I found changes in [docker-compose](https://github.com/moira-alert/moira/pull/811/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R12), maybe it is only for local dev and you forgot to revert

Tetrergeru:
> I found changes in [docker-compose](https://github.com/moira-alert/moira/pull/811/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R12), maybe it is only for local dev and you forgot to revert

Whoops, fixed

api/middleware/logger.go
```diff
@@ -87,56 +94,56 @@ func newLogEntry(logger moira.Logger, request *http.Request) *apiLoggerEntry {
...
-	entry.buf.WriteString("\n")
-	entry.buf.WriteString(string(stack))
-	log.Error(entry.buf.String())
+	entry.logger.Errorb().
+		Int("http_status", status).
+		Int("http_content_length", bytes).
+		Int("elapsed_time_ms", int(elapsed.Milliseconds())).
+		Interface("recovered_err", v).
+		String(moira.LogFieldNameStackTrace, string(stack)).
+		Msg(fmt.Sprintf("%s: panic", entry.msg))
```
kissken, line 146: Is panic msg arg or nessesary msg?

logging/zerolog_adapter/logger.go
```diff
@@ -71,84 +72,24 @@ func getLogWriter(logFileName string) (io.Writer, error) {
...
-	event.Timestamp().Msg(fmt.Sprint(args...))
-}
-
-func (l Logger) Infof(format string, args ...interface{}) {
-	event := l.Logger.Info()
-	if event == nil {
-		return
-	}
-	event.Timestamp().Msgf(format, args...)
+func (l Logger) Debugb() logging.EventBuilder {
```
kissken, line 75: hmm, but why here?

and may be now letter **b** in func name doesn't need? 

and maybe Logger struct don't need anymore?

at least I think now we should make Logger field in Logger struct make private to avoid using not our components?

kissken: - letter **b** will remove in another PR
- Logger and builder will help us to mock and test
- todo: make field private

Tetrergeru: Made field private

kissken (approved):

Dimedrolity (approved):

PR #817, opened 2022-12-16
Validate trigger target on create/update

Dimedrolity:
There is a problem that client sides should call /trigger/check/ API for validation, but obviously it is not necessary. As a result we have invalid triggers in DB. The plan consists of next steps:

1. Test current createTrigger and updateTrigger API.
2. Add new version of createTrigger and updateTrigger API with validation and test it. New version should not affect prev version, It is important for backward compatibility, because Moira is follows API first approach.

Then in another PRs update client sides to use new version of APIs.

### DIFF
--- a/api/dto/target.go
+++ b/api/dto/target.go
@@ -154,9 +154,9 @@ func TargetVerification(targets []string, ttl time.Duration, isRemote bool) []Tr
 	return functionsOfTargets
 }
 
-// HaveTreesError checks that at least one node of tree has a problem with error type.
+// DoesAnyTreeHaveError checks that at least one node of tree has a problem with error type.
 // It is wrapper to handle slice of trees.
-func HaveTreesError(trees []TreeOfProblems) bool {
+func DoesAnyTreeHaveError(trees []TreeOfProblems) bool {
 	for _, tree := range trees {
 		if tree.TreeOfProblems.hasError() {
 			return true
--- a/api/dto/target.go
+++ b/api/dto/target.go
@@ -17,8 +17,8 @@ func init() {
 type typeOfProblem string
 
 const (
-	IsWarn typeOfProblem = "warn"
-	IsBad  typeOfProblem = "bad"
+	isWarn typeOfProblem = "warn"
+	isBad  typeOfProblem = "bad"
 )
 
 var (
@@ -156,7 +156,7 @@ func checkExpression(expression parser.Expr, ttl time.Duration, isRemote bool) *
 
 		problemFunction.Problems = append(problemFunction.Problems, ProblemOfTarget{
 			Argument: argument,
-			Type:     IsBad,
+			Type:     isBad,
 			Position: 1,
 			Description: fmt.Sprintf(
 				"The function %s has a time sampling parameter %s larger than allowed by the config:%s",
@@ -187,7 +187,7 @@ func checkFunction(funcName string, isRemote bool) *ProblemOfTarget {
 	if _, isUnstable := unstableFunctions[funcName]; isUnstable {
 		return &ProblemOfTarget{
 			Argument:    funcName,
-			Type:        IsBad,
+			Type:        isBad,
 			Description: "This function is unstable: it can return different historical values with each evaluation. Moira will show unexpected values that you don't see on your graphs.",
 		}
 	}
@@ -195,7 +195,7 @@ func checkFunction(funcName string, isRemote bool) *ProblemOfTarget {
 	if _, isFalseNotification := falseNotificationsFunctions[funcName]; isFalseNotification {
 		return &ProblemOfTarget{
 			Argument:    funcName,
-			Type:        IsWarn,
+			Type:        isWarn,
 			Description: "This function shows and hides entire metric series based on their values. Moira will send frequent false NODATA notifications.",
 		}
 	}
@@ -203,7 +203,7 @@ func checkFunction(funcName string, isRemote bool) *ProblemOfTarget {
 	if _, isVisual := visualFunctions[funcName]; isVisual {
 		return &ProblemOfTarget{
 			Argument:    funcName,
-			Type:        IsWarn,
+			Type:        isWarn,
 			Description: "This function affects only visual graph representation. It is meaningless in Moira.",
 		}
 	}
@@ -211,7 +211,7 @@ func checkFunction(funcName string, isRemote bool) *ProblemOfTarget {
 	if !isRemote && !funcIsSupported(funcName) {
 		return &ProblemOfTarget{
 			Argument:    funcName,
-			Type:        IsBad,
+			Type:        isBad,
 			Description: "Function is not supported, if you want to use it, switch to remote",
 			Position:    0,
 			Problems:    nil,
@@ -273,7 +273,7 @@ func AreTreesHaveError(trees []TreeOfProblems) bool {
 }
 
 func isTreeHasError(tree *ProblemOfTarget) bool {
-	if tree.Type == IsBad {
+	if tree.Type == isBad {
 		return true
 	}
 
--- a/api/dto/target.go
+++ b/api/dto/target.go
@@ -100,17 +100,17 @@ var (
 	}
 )
 
-type problemOfTarget struct {
+type ProblemOfTarget struct {
 	Argument    string            `json:"argument"`
 	Type        typeOfProblem     `json:"type,omitempty"`
 	Description string            `json:"description,omitempty"`
 	Position    int               `json:"position"`
-	Problems    []problemOfTarget `json:"problems,omitempty"`
+	Problems    []ProblemOfTarget `json:"problems,omitempty"`
 }
 
 type TreeOfProblems struct {
 	SyntaxOk       bool             `json:"syntax_ok"`
-	TreeOfProblems *problemOfTarget `json:"tree_of_problems,omitempty"`
+	TreeOfProblems *ProblemOfTarget `json:"tree_of_problems,omitempty"`
 }
 
 // TargetVerification validates trigger targets.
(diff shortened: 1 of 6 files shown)


Dimedrolity:
Converted to draft because need to add validation on updating existing trigger `PUT trigger/<id>`

api/handler/trigger_test.go
```diff
@@ -95,3 +101,295 @@ func TestGetTrigger(t *testing.T) {
...
+
+			Convey("should return 200 and tree of problems", func() {
+				response := responseWriter.Result()
+				defer response.Body.Close()
+
+				So(response.StatusCode, ShouldEqual, http.StatusOK)
+
+				contentBytes, _ := io.ReadAll(response.Body)
+				contents := string(contentBytes)
+				fmt.Println(contents)
```
kissken, line 274: ops

api/dto/target.go
```diff
@@ -100,17 +100,17 @@ var (
 	}
 )
 
-type problemOfTarget struct {
+type ProblemOfTarget struct {
```
kissken, line 103: hmm, why become struct as public?

api/dto/target.go
```diff
@@ -261,3 +261,27 @@ func positiveDuration(argument parser.Expr) (string, time.Duration) {
 
 	return value, secondTimeDuration
 }
+
+func AreTreesHaveError(trees []TreeOfProblems) bool {
```
kissken, line 265: HaveTreesError - maybe

and missed godoc (

kissken: and i think this method will be better as method of struct

api/dto/triggers.go
```diff
@@ -377,30 +377,31 @@ type TriggerMaintenance struct {
...
+func (*ThrottlingResponse) Render(http.ResponseWriter, *http.Request) error {
 	return nil
 }
 
 type SaveTriggerResponse struct {
-	ID      string `json:"id"`
-	Message string `json:"message"`
+	ID      string               `json:"id"`
+	Message string               `json:"message"`
+	Check   TriggerCheckResponse `json:"check,omitempty"`
```
kissken, line 395: here will be errors, yea?

api/dto/target.go
Dimedrolity: This struct is used in tests when compare expected and actual

api/handler/trigger_test.go
Dimedrolity: Thanks, fixed https://github.com/moira-alert/moira/pull/817/commits/156f3b83d9e2d5f2279a08abbca0dccdb5a3d9eb

api/dto/triggers.go
Dimedrolity: Yes, it is response struct for create/update APIs. Now it contains result of validation with tree of problems, and each problem has type (yellow/red). So `Check` field is result of existing /check/ API

kissken: not sure with naming, maybe as in "check" named "check_result", but nothing is in my mind better 
maybe Oleg suggests smth

api/dto/target.go
kissken: but this method is still not attached to the struct :)

api/dto/target.go
kissken: only for tests? I think better leave struct private to avoid unexpected usage and make constructor....

api/dto/triggers.go
Tetrergeru: `check_result` seems fine. Simple `check` seems a little to general

api/handler/trigger.go
```diff
@@ -39,19 +39,62 @@ func updateTrigger(writer http.ResponseWriter, request *http.Request) {
...
+		response.Check.Targets = problems
+	}
+
 	if err := render.Render(writer, request, response); err != nil {
 		render.Render(writer, request, api.ErrorRender(err)) //nolint
 		return
 	}
 }
 
+func isNeedValidate(request *http.Request) bool {
```
Tetrergeru, line 68: Maybe we should rename function to be a correct english phrase

```suggestion
func needsValidation(request *http.Request) bool {
```

api/dto/triggers.go
Dimedrolity: Well, I will rename to `checkResult` (JSON is camelCase)

api/dto/target.go
```diff
@@ -140,8 +154,20 @@ func TargetVerification(targets []string, ttl time.Duration, isRemote bool) []Tr
 	return functionsOfTargets
 }
 
+// HaveTreesError checks that at least one node of tree has a problem with error type.
+// It is wrapper to handle slice of trees.
+func HaveTreesError(trees []TreeOfProblems) bool {
```
Tetrergeru, line 159: Mabe this is better?

```suggestion
func DoTreesHaveErrors(trees []TreeOfProblems) bool {
```

Tetrergeru (changes requested):
Left some suggestions

api/dto/triggers.go
Dimedrolity: Renamed field and JSON key, https://github.com/moira-alert/moira/pull/817/commits/10f6c6e91319b1b768ddf8c53cb9242cc7283f94

api/handler/trigger.go
Dimedrolity: Following english language rules makes code too perfect, I think it is redundant. There are well naming conventions to use `is, has, can`. Also I think `needValidate` is enough, but `is` prefix is well to stress that it is predicate

api/dto/target.go
Dimedrolity: `DoTreesHaveError` is fine, not error**s**, it is singular. Maybe it is need to rename to `isSomeTreeHasError` to stress that it checks that at least one error in all trees

api/handler/trigger.go
Tetrergeru: I think, that predicates should look good, when they are put in condition statement like `if needsValidation(request) { ... }`.

The return type of bool already stresses that this function is a predicate. We already have functions with similar semantics that don't follow this convention (e.g. [this](https://github.com/moira-alert/moira/blob/759d70687dc7ed25768b202a4af177c1744c67cd/checker/worker/metrics.go#L27))

Dimedrolity: > I think, that predicates should look good, when they are put in condition statement

I agree with this approach, then `if needValidate(request)` is the best variant, it is easy to read. In your example `needHandlePattern` we have `need`, not `needs`, there is no `needs` in Moira

Tetrergeru: Ok, then let's have `needValidate`

api/dto/target.go
Tetrergeru: I think `do` stresses the fact that it is a predicate well enough. Let's name it `DoesAnyTreeHaveError`. `Any` is someting commonly used in predicates, so it will be easy to read

api/handler/trigger.go
Dimedrolity: Well, I agree with `needValidate`. @kissken do you agree?

api/dto/target.go
Dimedrolity: `DoesAnyTreeHaveError` is nice, @kissken do you agree?

api/handler/triggers.go
```diff
@@ -51,13 +51,23 @@ func getAllTriggers(writer http.ResponseWriter, request *http.Request) {
...
 func createTrigger(writer http.ResponseWriter, request *http.Request) {
 	trigger, err := getTriggerFromRequest(request)
 	if err != nil {
 		render.Render(writer, request, err) //nolint
 		return
 	}
 
+	var problems []dto.TreeOfProblems
+	if isNeedValidate(request) {
+		problems = validateTargets(request, trigger)
```
kissken, line 64: why dont use := ? :)

api/handler/triggers.go
```diff
@@ -74,6 +84,10 @@ func createTrigger(writer http.ResponseWriter, request *http.Request) {
 		return
 	}
 
+	if problems != nil {
+		response.CheckResult.Targets = problems
```
kissken, line 88: hmm, in writeErrorSaveResponse already init?

api/handler/trigger.go
kissken: yep

api/dto/target.go
kissken: pff, ok

api/handler/trigger.go
Dimedrolity: Done https://github.com/moira-alert/moira/pull/817/commits/7c68d2f125f4275b5d23fc8573896cbc67fc3f9c

api/dto/target.go
Dimedrolity: Done https://github.com/moira-alert/moira/pull/817/commits/7c68d2f125f4275b5d23fc8573896cbc67fc3f9c

api/handler/triggers.go
Dimedrolity: Can't use `:=` on 64th line, because need to write problems to reponse on 88th line. Can't use `:=` on 62th line, because need to validate only if `needValidate`

api/dto/target.go
kissken: i think myself that public structs are evil, but ok, lets leave, and add godoc please)

api/handler/triggers.go
Dimedrolity: `writeErrorSaveResponse` is only for error (red color functions), and there is `return` [on 47th line](https://github.com/moira-alert/moira/pull/817/files#diff-a99418ecdfd4ca5c347514b729fa53b494213b93e43328a6a753a32238d7abe1R47). Here is on 88th line writing problems to response for warn (yellow color functions), so it saves trigger but writes problems to response body

kissken: hm, i think after all we should move it into save func....

kissken (approved):

Tetrergeru (dismissed):

kissken (approved):

Tetrergeru (approved):

PR #825, opened 2023-02-10
fix(filter): fix regex in series by tag parser

### LINKED ISSUE #823: Dont work regex groups in target
# BUG REPORT

## What did you expect to see?
If i wrote in some tag (for tagged metric) like "!=~^.*(a|b|c).*$" 
No metric's is written in metric-three and trigger always has been in no_data state

## What did you see instead?

We want use regex group in tagged metric

balalay12:
# PR Summary
This PR fix `series by tags` parser. 
1. Fix regular expression for correct work with regular expressions in tags. Example: `seriesByTag("respCode=~^(4|5)\d{2}")`
2. Now transform function applied to every tag spec. Not for whole value in `seriesByTag()`.
3. If spec was tranformed change `operator` from `equal` to `match`.

Closes #823

### DIFF
--- a/filter/series_by_tag.go
+++ b/filter/series_by_tag.go
@@ -39,7 +39,7 @@ type TagSpec struct {
 
 func transformWildcardToRegexpInSeriesByTag(input string) (string, bool) {
 	var (
-		result                                     = input
+		result := input
 		correctLengthOfMatchedWildcardIndexesSlice = 4
 		isTransformed                              = false
 	)
--- a/filter/series_by_tag.go
+++ b/filter/series_by_tag.go
@@ -6,9 +6,12 @@ import (
 	"strings"
 )
 
-var tagSpecRegex = regexp.MustCompile(`^["']([^,!=]+)\s*(!?=~?)\s*([^,]*)["']`)
-var tagSpecDelimiterRegex = regexp.MustCompile(`^\s*,\s*`)
-var seriesByTagRegex = regexp.MustCompile(`^seriesByTag\(([^)]+)\)$`)
+var (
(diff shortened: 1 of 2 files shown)


kissken:
before merge we will test it on feature branch :)

filter/series_by_tag.go
```diff
@@ -34,13 +37,15 @@ type TagSpec struct {
 	Value    string
 }
 
-func transformWildcardToRegexpInSeriesByTag(input string) string {
-	var result = input
-	var re = regexp.MustCompile(`\{(.*?)\}`)
-	var correctLengthOfMatchedWildcardIndexesSlice = 4
+func transformWildcardToRegexpInSeriesByTag(input string) (string, bool) {
+	var (
+		result                                     = input
```
Tetrergeru, line 42: I think it would be more ideomatic to use `:=` instead of the var bloc, if it is possible
```suggestion
		result := input
```

filter/series_by_tag.go
```diff
@@ -34,13 +37,15 @@ type TagSpec struct {
...
+		correctLengthOfMatchedWildcardIndexesSlice = 4
```
Tetrergeru, line 43: Maybe we sould make it a constant and also use in `ParseSeriesByTag`, where magic number 4 is used

Tetrergeru (changes requested):
Left some minor suggestions

Tetrergeru (dismissed):

kissken:
потестили, результаты:

seriesByTag("name=~^a.(b|c).d") ✅
seriesByTag('name=~^a.(b|c).d') ❌
seriesByTag('name=a.b.c')❌

kissken:
последние правки были ок

kissken:
влила мастер в ветку

kissken (approved):

Tetrergeru (approved):

PR #841, opened 2023-04-10
Add more logs and move tags deletion

kissken:
# PR Summary

make logs more informative and move tags deletion out of last checks deletion

all changes in tests - because of linter :)

Closes/Relates #issue


checker/fetch_test.go
```diff
@@ -7,17 +7,17 @@ import (
 	"github.com/golang/mock/gomock"
 	"github.com/moira-alert/moira"
 	metricSource "github.com/moira-alert/moira/metric_source"
-	mock_metric_source "github.com/moira-alert/moira/mock/metric_source"
-	mock_moira_alert "github.com/moira-alert/moira/mock/moira-alert"
+	mockmetricsource "github.com/moira-alert/moira/mock/metric_source"
```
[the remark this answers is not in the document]
kissken: ugu
image

cmd/cli/user.go
```diff
@@ -108,11 +108,19 @@ func usersCleanup(logger moira.Logger, database moira.Database, users []string,
 		usersMap[user] = true
 	}
 
+	logger.Info().
+		Int("count", len(usersMap)).
+		Msg("actual users")
```
Tetrergeru, line 113: I think we use capital leters, since it's a complete sentence
```suggestion
		Msg("Actual users")
```

database/redis/metric.go
```diff
@@ -306,16 +306,17 @@ func (connector *DbConnector) RemoveMetricRetention(metric string) error {
 }
 
 // RemoveMetricValues remove metric timestamps values from 0 to given time
-func (connector *DbConnector) RemoveMetricValues(metric string, toTime int64) error {
+func (connector *DbConnector) RemoveMetricValues(metric string, toTime int64) (int64, error) {
```
Tetrergeru, line 309: Maybe we should use named return parameter, so that it would be more obvious what does this int64 mean

kissken: do u think is it ok use in one func in all pkg?

Tetrergeru (approved):

PR #847, opened 2023-05-30
Add support for sentinel auth

DreidSpb:
Added new configuration option sentinel_pass and passed it to redis library.

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -27,6 +27,8 @@ type RedisConfig struct {
 	Addrs string `yaml:"addrs"`
 	// Redis Sentinel password
 	SentinelPassword string `yaml:"sentinel_password"`
+	// Redis Sentinel username
+	SentinelUsername string `yaml:"sentinel_username"`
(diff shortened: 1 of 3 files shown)


cmd/config.go
```diff
@@ -25,6 +25,8 @@ type RedisConfig struct {
 	MasterName string `yaml:"master_name"`
 	// Redis address list, format: {host1_name:port},{ip:port}
 	Addrs string `yaml:"addrs"`
+	// Redis Sentinel password
+	SentinelPassword string `yaml:"sentinel_password"`
```
kissken, line 29: May be should we add SentinelUser too? :)

DreidSpb: I am sorry, sure we should.
I made this patch for my own deployment, we don't use username in our sentinel.
And I forgot to add it because thought to make a pull request hit me later :)
Will make another commit right now.

Tetrergeru (dismissed):

PR #875, opened 2023-07-04
fix(checker): fix transition to ERROR state when duplicating metrics

almostinf:
fix transition to ERROR state when duplicating metrics

Replaced the ERROR condition with EXCEPTION for duplicate names in `checker/check.go`, fixed the corresponding tests

Also fixed 2 linter errors in `database/redis` and `senders/pagerduty`

### DIFF
--- a/senders/pagerduty/send.go
+++ b/senders/pagerduty/send.go
@@ -18,7 +18,7 @@ const summaryMaxChars = 1024
 // SendEvents implements Sender interface Send
 func (sender *Sender) SendEvents(events moira.NotificationEvents, contact moira.ContactData, trigger moira.TriggerData, plots [][]byte, throttled bool) error {
 	event := sender.buildEvent(events, contact, trigger, plots, throttled)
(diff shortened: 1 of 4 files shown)


senders/pagerduty/send.go
```diff
@@ -17,7 +18,7 @@ const summaryMaxChars = 1024
 // SendEvents implements Sender interface Send
 func (sender *Sender) SendEvents(events moira.NotificationEvents, contact moira.ContactData, trigger moira.TriggerData, plots [][]byte, throttled bool) error {
 	event := sender.buildEvent(events, contact, trigger, plots, throttled)
-	_, err := pagerduty.ManageEvent(event)
+	_, err := pagerduty.ManageEventWithContext(context.TODO(), event)
```
Tetrergeru, line 21: I think we shouldn't merge `context.TODO()` in master.

Let's use `context.Backgound()` here for the time being

almostinf: Fixed

Tetrergeru (approved):

PR #876, opened 2023-07-08
feat: support sprig library for trigger templates

gODeaLoAple:
# Add sprig library templates

https://pkg.go.dev/github.com/masterminds/sprig#section-readme

Also filter some functions in sprig to prevent possible unsafety/insecure usages


templating/templating.go
```diff
@@ -43,6 +44,19 @@ type trigger struct {
 	Name string `json:"name"`
 }
 
+type FuncMap = template.FuncMap
```
kissken, line 47: new type is extra, i think

templating/templating.go
```diff
@@ -64,12 +78,198 @@ func Populate(name, description string, events []Event) (desc string, err error)
 		"stringsSplit":      strings.Split,
 	}
 
+	sprigFuncMap := FilterKeys(sprig.FuncMap(), []string {
```
kissken, line 81: better call FilterKeys only once for start, not with call "Populate"

gODeaLoAple: I moved out this and funcsMap out of "Populate" scope

templating/templating.go
```diff
@@ -43,6 +45,214 @@ type trigger struct {
 	Name string `json:"name"`
 }
 
+func FilterKeys(source template.FuncMap, keys []string) template.FuncMap {
+	result := template.FuncMap{}
+	for _, key := range keys {
+		value, found := source[key]
+		if found {
```
Tetrergeru, line 52:
```suggestion
		if value, ok := source[key], ok {
```

Tetrergeru (approved):

templating/templating.go
```diff
@@ -43,6 +45,214 @@ type trigger struct {
 	Name string `json:"name"`
 }
 
+func FilterKeys(source template.FuncMap, keys []string) template.FuncMap {
```
kissken, line 48: and let's fix linter and add godoc for this func or make func private better FilterKeys=> filterKeys :)

templating/templating.go
gODeaLoAple: Does not work for me
image

templating/templating.go
gODeaLoAple: done

templating/templating.go
kissken: if value, ok := source[key], ok { => if value, ok := source[key]**;** ok {

kissken: not ,
use ; in this :)

Tetrergeru: Sorry, it was a typo

Tetrergeru (approved):

templating/templating.go
gODeaLoAple: fixed

kissken (approved):

PR #877, opened 2023-07-11
refactor(senders): refactor init in senders with mapstructure

almostinf:
# Refactor init in senders

Rewrote the method of getting senders from yaml config using `mapstructure` package

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -47,6 +47,8 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 )
 
+require github.com/mitchellh/mapstructure v1.5.0
+
 require (
 	bitbucket.org/tebeka/strftime v0.0.0-20140926081919-2194253a23c0 // indirect
 	github.com/JaderDias/movingmedian v0.0.0-20220813210630-d8c6b6de8835 // indirect
@@ -119,7 +121,6 @@ require (
 	github.com/minio/md5-simd v1.1.2 // indirect
 	github.com/minio/minio-go/v7 v7.0.45 // indirect
 	github.com/minio/sha256-simd v1.0.0 // indirect
-	github.com/mitchellh/mapstructure v1.5.0 // indirect
 	github.com/mjibson/go-dsp v0.0.0-20180508042940-11479a337f12 // indirect
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.2 // indirect
--- a/senders/discord/init.go
+++ b/senders/discord/init.go
@@ -35,12 +35,12 @@ type Sender struct {
 
 // Init reads the yaml config
 func (sender *Sender) Init(senderSettings map[string]interface{}, logger moira.Logger, location *time.Location, dateTimeFormat string) error {
-	var discord discord
-	err := mapstructure.Decode(senderSettings, &discord)
+	var ds discord
+	err := mapstructure.Decode(senderSettings, &ds)
 	if err != nil {
 		return fmt.Errorf("failed to decode senderSettings to discord config: %w", err)
 	}
-	token := discord.Token
+	token := ds.Token
 	if token == "" {
 		return fmt.Errorf("cannot read the discord token from the config")
 	}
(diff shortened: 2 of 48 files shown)


almostinf:
I can additionally add custom-headers to the webhook tests

senders/discord/init.go
```diff
@@ -16,6 +17,12 @@ const (
 	workerName      = "DiscordBot"
 )
 
+// Structure that represents the Discord configuration in the YAML file
+type Discord struct {
```
Tetrergeru, line 21: We can make this type (and other sender configuration types) private, since it is only used in this package

senders/discord/init.go
```diff
@@ -27,9 +34,13 @@ type Sender struct {
...
 
 // Init reads the yaml config
-func (sender *Sender) Init(senderSettings map[string]string, logger moira.Logger, location *time.Location, dateTimeFormat string) error {
-	var err error
-	token := senderSettings["token"]
+func (sender *Sender) Init(senderSettings map[string]interface{}, logger moira.Logger, location *time.Location, dateTimeFormat string) error {
+	var discord Discord
+	err := mapstructure.Decode(senderSettings, &discord)
+	if err != nil {
+		return fmt.Errorf("decoding error from yaml file to discord structure: %s", err)
```
Tetrergeru, line 41:
```suggestion
		return fmt.Errorf("failed to decode senderSettings to config: %w", err)
```

Tetrergeru: And similar for other senders

senders/pagerduty/init.go
```diff
@@ -19,8 +20,12 @@ type Sender struct {
 }
 
 // Init loads yaml config, configures the pagerduty client
-func (sender *Sender) Init(senderSettings map[string]string, logger moira.Logger, location *time.Location, dateTimeFormat string) error {
-	sender.frontURI = senderSettings["front_uri"]
+func (sender *Sender) Init(senderSettings map[string]interface{}, logger moira.Logger, location *time.Location, dateTimeFormat string) error {
+	var ok bool
+	sender.frontURI, ok = senderSettings["front_uri"].(string)
```
Tetrergeru, line 25: Lets use config struct as in other sender for consistency

senders/twilio/twilio.go
```diff
@@ -18,32 +19,48 @@ type sendEventsTwilio interface {
...
+	var twilio Twilio
+	err := mapstructure.Decode(senderSettings, &twilio)
+	if err != nil {
+		return fmt.Errorf("decoding error from yaml file to twilio structure: %s", err)
+	}
+	apiType := twilio.Type
 
-	apiASID := senderSettings["api_asid"]
+	apiASID := twilio.APIAsid
 	if apiASID == "" {
```
Tetrergeru, line 49: Do we need these check if we already have `mapstructure.Decode`?

senders/webhook/webhook.go
```diff
@@ -7,9 +7,20 @@ import (
...
 	"github.com/moira-alert/moira"
 )
 
+// Structure that represents the Webhook configuration in the YAML file
+type WebHook struct {
+	Name          string              `mapstructure:"name"`
+	URL           string              `mapstructure:"url"`
+	User          string              `mapstructure:"user"`
+	Password      string              `mapstructure:"password"`
+	CustomHeaders []map[string]string `mapstructure:"custom-headers"`
```
Tetrergeru, line 20: Maybe like this?
```suggestion
	CustomHeaders []map[string]string `mapstructure:"custom-headers,omitempty"`
```

go.mod
```diff
@@ -47,6 +47,8 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 )
```
Tetrergeru, line 49: Please rename the PR according to git commit [style guide](https://github.com/moira-alert/moira/blob/master/.github/CONTRIBUTING.md#git-style-guide)

senders/discord/init.go
almostinf: fixed

senders/twilio/twilio.go
almostinf: I think that all checks are needed, because by default mapstructure leaves zero value if the field is not filled

senders/webhook/webhook.go
almostinf: fixed, but it will have the same result with or without omitempty

go.mod
almostinf: fixed

PR #883, opened 2023-07-24
fix(senders): fix text in sms and email

almostinf:
# Fix text in sms and email

Refactored the GetMetricsValues function to be able to reduce the message length

### DIFF
--- a/datatypes.go
+++ b/datatypes.go
@@ -32,6 +32,13 @@ const (
 	limit         = 1000
 )
 
+type NotificationEventSettings int
+
+const (
+	None NotificationEventSettings = iota
+	ShortMessage
+)
+
 // NotificationEvent represents trigger state changes event
 type NotificationEvent struct {
 	IsTriggerEvent   bool               `json:"trigger_event,omitempty"`
@@ -411,7 +418,7 @@ func (notification *ScheduledNotification) GetKey() string {
 		notification.Event.Metric,
 		notification.Event.State,
 		notification.Event.Timestamp,
-		notification.Event.GetMetricsValues(false),
+		notification.Event.GetMetricsValues(None),
 		notification.SendFail,
 		notification.Throttled,
 		notification.Timestamp,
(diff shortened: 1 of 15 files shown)


datatypes.go
```diff
@@ -447,32 +449,49 @@ func (schedule *ScheduleData) IsScheduleAllows(ts int64) bool {
...
 
 func (event NotificationEvent) String() string {
-	return fmt.Sprintf("TriggerId: %s, Metric: %s, Values: %s, OldState: %s, State: %s, Message: '%s', Timestamp: %v", event.TriggerID, event.Metric, event.GetMetricsValues(), event.OldState, event.State, event.CreateMessage(nil), event.Timestamp)
+	return fmt.Sprintf("TriggerId: %s, Metric: %s, Values: %s, OldState: %s, State: %s, Message: '%s', Timestamp: %v", event.TriggerID, event.Metric, event.GetMetricsValues(false), event.OldState, event.State, event.CreateMessage(nil), event.Timestamp)
 }
 
 // GetMetricsValues gets event metric value and format it to human readable presentation
-func (event NotificationEvent) GetMetricsValues() string {
-	var targetNames []string //nolint
+func (event NotificationEvent) GetMetricsValues(isLess bool) string {
```
Tetrergeru, line 456: `isLess` — maybe we should rename it somehow? Now it's too difficult to understand the purpose of the variable from it's name.

Maybe we sould introduse an enumeration instead of a boolean flag, so that it would be more readable and could be expanded in the future

almostinf: I agree, I've refactored it now

datatypes.go
```diff
@@ -28,6 +29,14 @@ const (
...
 	format        = "15:04 02.01.2006"
 	remindMessage = "This metric has been in bad state for more than %v hours - please, fix."
+	limit         = 1000
+)
+
+type notificationEventSettings int
+
+const (
+	DefaultNotificationSettings notificationEventSettings = iota
+	ShortMessage
```
Tetrergeru, line 39: Maybe it would be more clear if we name it like that?

```suggestion
	SIFormatNumbers
```

almostinf: Ok, I've fixed it

kissken (approved):

PR #891, opened 2023-08-05
feature(api): add swagger specification generation

almostinf:
# Add swagger specification generation

I used the [swaggo](https://github.com/swaggo/swag) package to generate the specification, for which I added annotations over all handlers and examples in dto models. I also added handy commands to work with swagger specification in `Makefile`

The documentation is available at the url path:
`/api/swagger/index.html`


Makefile
```diff
@@ -39,6 +39,15 @@ lint:
 mock:
 	. ./generate_mocks.sh
 
+.PHONY: install-swag
+install-swag:
+	go install github.com/swaggo/swag/cmd/swag@latest
```
kissken, line 44: ток не латест:)

almostinf: поправил)

api/handler/config.go
```diff
@@ -2,6 +2,24 @@ package handler
 
 import "net/http"
 
+type Contact struct {
+	Type  string `json:"type" example:"telegram"`
```
Tetrergeru, line 6: Maybe we should prefix these types with `Example` so that nobody meistake them for actual types used in code?

Tetrergeru: Or maybe change code, so that it uses actual types instead of passing `[]byte` through the entirety of the call stack? Maybe not in this PR, but later — let's make a task

api/handler/contact.go
```diff
@@ -27,6 +27,16 @@ func contact(router chi.Router) {
 	})
 }
 
+// nolint: gofmt,goimports
+//
+//	@summary	Gets all Moira contacts
+//	@id			get-all-contacts
+//	@tags		contact
+//	@produce	json
+//	@success	200	{object}	dto.ContactList					"Contacts fetched successfully"
```
Tetrergeru, line 36: Tabs look broken. It's okay or should we raplcae them with spaces?

datatypes.go
```diff
@@ -42,16 +42,16 @@ const (
 
 // NotificationEvent represents trigger state changes event
 type NotificationEvent struct {
-	IsTriggerEvent   bool               `json:"trigger_event,omitempty"`
-	Timestamp        int64              `json:"timestamp"`
-	Metric           string             `json:"metric"`
-	Value            *float64           `json:"value,omitempty"`
+	IsTriggerEvent   bool               `json:"trigger_event,omitempty" example:"true"`
+	Timestamp        int64              `json:"timestamp" example:"1590741878"`
+	Metric           string             `json:"metric" example:"carbon.agents.*.metricsReceived"`
```
Tetrergeru, line 47: Can this field actually contain a `*` character?

api/handler/config.go
almostinf: Okay, I've renamed it, I'll do the task soon

api/handler/contact.go
almostinf: They are crookedly displayed in PR, if you do a pull of my branch you'll see that the tabs are normal there, that's why `swag fmt` is called after generating the documentation

datatypes.go
almostinf: This strange field used to be used before when sending sms before I fixed the problem

Tetrergeru (approved):

PR #892, opened 2023-08-07
feat(api): added full trigger validation in api

lordvidex:
# PR Summary
Trigger is fully validated using [`expr`](https://github.com/antonmedv/expr) when trigger is created or updated.

## Problem solved
It was possible for `govalute` to skip validations for expressions that do not fall into the current conditional branch based on the target values.
For example, `t1 > 10 ? ok : (t2 * 10 : error ? warn)` would return result without error where `t1 > 10` without considering the syntax error (misplacement of ternary operators) in the expression.

## Notes
Commit https://github.com/moira-alert/moira/commit/7bb6f5af4c68bf4410663c5954f0834e391591c7 removes expression optimization because the `vm.Program` in expr caches the result of Get() function which may produce wrong result in `expr.Run` function. Check benchmark and commit description for more.

## Possible problems
- `expr` is not fully compatible with `govalute` 
     - expr supports functions
     - expr uses `[a, b, c,...]` while govalute uses `(a,b,c,...)` for array literals
    
Because of the following errors, `expr` might mark an expression as valid where `govalute` will return errors and vice versa.
 
## Relations
Augments #881


expression/expression.go
```diff
@@ -95,7 +113,49 @@ func (triggerExpression *TriggerExpression) Evaluate() (moira.State, error) {
 	}
 }
 
-func validateUserExpression(triggerExpression *TriggerExpression, userExpression *govaluate.EvaluableExpression) (*govaluate.EvaluableExpression, error) {
+// Validate gets trigger expression and validates it for given parameters using expr
+func (triggerExpression *TriggerExpression) Validate() error {
+	expression, err := getExpression(triggerExpression)
```
almostinf, line 118: Right now your solution is doing validation even for triggers from simple mode, maybe you should make it do it only for user expressions?
I suggest this change: you can leave the `triggerExpression.Evaluate()` call, but change the `validateUserExpression` function

Then in essence there will be 2 steps of running the expression:
- in `expr` the compilation stage
- in `govaluate` the stage of program execution

expression/expression.go
```diff
@@ -95,7 +113,49 @@ func (triggerExpression *TriggerExpression) Evaluate() (moira.State, error) {
...
+			expression.String(),
+			expr.Patch(triggerExpression), // patch identifiers with call to Get
+			expr.Optimize(true),
+		)
+		if err != nil {
+			return err
+		}
+		exprCache.Set(cacheKey, program, cache.NoExpiration)
+	}
+	result, err := expr.Run(program, map[string]interface{}{
```
almostinf, line 141: Should we do a compilation here? Let the faster `govaluate` do the execution of the expression

expression/expression.go
```diff
@@ -115,7 +175,7 @@ func getExpression(triggerExpression *TriggerExpression) (*govaluate.EvaluableEx
 			return nil, err
 		}
 
-		return validateUserExpression(triggerExpression, userExpression)
+		return validateExpressionVariables(triggerExpression, userExpression)
```
almostinf, line 178: After you patched expr by replacing them with Get, expr won't run all the arguments through Get? Because if it does, then there is no point in validating the arguments a second time

expression/expression_test.go
```diff
@@ -19,82 +20,137 @@ func TestExpression(t *testing.T) {
 	Convey("Test Default", t, func() {
 		warnValue := 60.0
 		errorValue := 90.0
-		result, err := (&TriggerExpression{MainTargetValue: 10.0, WarnValue: &warnValue, ErrorValue: &errorValue, TriggerType: moira.RisingTrigger}).Evaluate()
-		So(err, ShouldBeNil)
+		expression := &TriggerExpression{MainTargetValue: 10.0, WarnValue: &warnValue, ErrorValue: &errorValue, TriggerType: moira.RisingTrigger}
```
almostinf, line 23: As if you could not validate an expression from Simple Mode, because it will always be correctly written, the problem may be only in the calculation, but this is already a `gavaluate` task

expression/expression_test.go
```diff
@@ -19,82 +20,137 @@ func TestExpression(t *testing.T) {
...
+		result, err1 = trigger.Evaluate()
+		err2 = trigger.Validate()
+		So(err1, ShouldBeNil)
+		So(err2, ShouldBeNil)
 		So(result, ShouldResemble, moira.StateNODATA)
 
 		expression = "t1 > 10 && t2 > 3 ? OK : ddd"
-		result, err = (&TriggerExpression{Expression: &expression, MainTargetValue: 11.0, AdditionalTargetsValues: map[string]float64{"t2": 4.0}, TriggerType: moira.ExpressionTrigger}).Evaluate()
-		So(err, ShouldResemble, ErrInvalidExpression{fmt.Errorf("invalid variable value: %w", fmt.Errorf("no value with name ddd"))})
+		trigger = &TriggerExpression{Expression: &expression, MainTargetValue: 11.0, AdditionalTargetsValues: map[string]float64{"t2": 4.0}, TriggerType: moira.ExpressionTrigger}
```
almostinf, line 133: The 2 errors look a bit odd, let's keep the one error that comes back from Evaluate, which will either be at runtime from `govaluate` or at compile time from `expr`

expression/expression_test.go
```diff
@@ -225,6 +281,58 @@ func TestGetExpressionValue(t *testing.T) {
...
+				},
+			}
+			err := triggerExpression.Validate()
+			t.Log(err)
+			So(err, ShouldNotBeNil)
+		}
+	})
+}
+
+func ptr[T any](obj T) *T {
```
almostinf, line 332: So, why do you need this function if all your expressions are strings? I mean of the same type

lordvidex: there were some tests that had pointers to float64 like `WarnValue` and `ErrorValue`. I think I didn't later use them, can I change it to `stringPtr`?

expression/expression.go
lordvidex: Okay, I'll make the following changes:
- remove the Validate() function
- the current actions in the Validate() function should be in the `validateUserExpression` function after checking for variables.
- this has to still be cached because we don't want to validate an expression that has already been validated before because checker calls `Evaluate` frequently.

Look at this and let me know if this is good to go @almostinf

expression/expression_test.go
almostinf: What I mean is that `ptr` in the code is only used for expressions that are always strings, you could just do as in `WarnValue` or `ErrorValue` and write:
```go
Expression: &expr,
```
Is not it?

expression/expression.go
lordvidex: there are two ways:
- if the `expr.Run()` is removed, then we will still need to check for the expression variables, because `expr` checks for this variables when Running.
- if the `expr.Run()` is kept, we can totally remove the variable checking logic.

I prefer the second option, what do you think about this?

expression/expression_test.go
lordvidex: I believe that this test was added to prevent breaking changes, i added the call to Validate() to make sure that simple expressions have a 100% guarantee to be safe as well. I can remove it if you want.

expression/expression.go
lordvidex: `Run` needs to be called for `expr` to check further for validation. I'll give an example:
the compile stage changes `t1 => Get("t1")` but it's during the execution that the error returned by Get("t1") will be returned if "t1" does not exist.

The compilation stage only checks for basic syntax errors such as misplacing ternary operators.

expression/expression.go
almostinf: If `expr.Run()` checks all variables at all, even those in other branches, then this is a good option and there is no need to recheck them

expression/expression_test.go
lordvidex: Ohh true, the function is truly useless here

expression/expression.go
almostinf: Yeah, it's a good plan:
- Move all your logic to the `validateUserExpression` function, thus not breaking the previous code, but adding only a new validation step
- Also, you won't need to change the previous tests much, `Evaluate()` will be called in them as well, you will just need to add more test cases
- Also, if `expr.Run()`, checks all arguments for validity, you don't need to validate them a second time

expression/expression.go
almostinf: If `expr.Run()` checks absolutely all arguments, no problem, you can leave it like this

expression/expression.go
lordvidex: This will however alter the expected error for some of the tests

almostinf: I don't think it's that scary

perfomance_tests/expression/expression_test.go
```diff
@@ -26,19 +31,52 @@ func BenchmarkDefault2Expr(b *testing.B) {
 		MainTargetValue: 10.0,
 		WarnValue:       &warnValue,
 		ErrorValue:      &errorValue,
+		TriggerType:     moira.RisingTrigger,
```
kissken, line 34: hmmm, why it become necessary?

lordvidex: I debugged the benchmark and it was actually returning errors, therefore the tests were not really testing the speed of expression parsing.

expression/expression.go
```diff
@@ -95,13 +98,62 @@ func (triggerExpression *TriggerExpression) Evaluate() (moira.State, error) {
...
 	}
-	return userExpression, nil
+
+	env := map[string]interface{}{
+		"ok":         moira.StateOK,
+		"error":      moira.StateERROR,
+		"warn":       moira.StateWARN,
+		"warning":    moira.StateWARN,
+		"nodata":     moira.StateNODATA,
+		"t1":         triggerExpression.MainTargetValue,
```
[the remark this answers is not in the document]
lordvidex: on :135 I ranged over the `triggerExpression.AdditionalTargetsValues` to add them to env.

expression/expression.go
```diff
@@ -95,13 +98,62 @@ func (triggerExpression *TriggerExpression) Evaluate() (moira.State, error) {
...
-func validateUserExpression(triggerExpression *TriggerExpression, userExpression *govaluate.EvaluableExpression) (*govaluate.EvaluableExpression, error) {
-	for _, v := range userExpression.Vars() {
-		if _, err := triggerExpression.Get(v); err != nil {
-			return nil, fmt.Errorf("invalid variable value: %w", err)
+// validateUserExpression uses expr package to validate <user defined> trigger expressions
+func validateUserExpression(
+	triggerExpression *TriggerExpression,
+	userExpression *govaluate.EvaluableExpression,
+) (*govaluate.EvaluableExpression, error) {
+	expression := *triggerExpression.Expression
```
almostinf, line 106: and there is `triggerExpression.Expression` will be nil?

expression/expression.go
```diff
@@ -95,13 +98,62 @@ func (triggerExpression *TriggerExpression) Evaluate() (moira.State, error) {
...
+	}
+	program, err := expr.Compile(
+		strings.ToLower(expression),
+		expr.Optimize(true),
+		expr.Env(env),
+	)
+	if err != nil {
+		return nil, err
+	}
+	result, err := expr.Run(program, env)
```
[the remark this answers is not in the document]
lordvidex: This is true. I'll test again to verify that `expr.Run` is not needed, then I'll remove it.

expression/expression.go
almostinf: This is a potentially dangerous place that could lead to panic if another developer calls this function elsewhere, it's best to protect yourself from such a thing as much as possible

lordvidex: I agree, I'll push a fix now

expression/expression.go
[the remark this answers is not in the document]
lordvidex: done.

expression/expression.go
```diff
@@ -77,6 +84,54 @@ func (triggerExpression TriggerExpression) Get(name string) (interface{}, error)
 	}
 }
 
+// Validate returns error if triggers of type moira.ExpressionTrigger are badly formatted otherwise nil
+func (triggerExpression *TriggerExpression) Validate() error {
+	if triggerExpression.TriggerType != moira.ExpressionTrigger {
+		return fmt.Errorf("only advanced triggers should be validated")
+	}
+	if triggerExpression.Expression == nil {
```
almostinf, line 92: It is better to leave the same check as in getExpression:
```go
if triggerExpression.Expression == nil || *triggerExpression.Expression == "" {
			return fmt.Errorf("trigger_type set to expression, but no expression provided")
}
```

expression/expression.go
```diff
@@ -28,6 +31,10 @@ func (err ErrInvalidExpression) Error() string {
 	return err.internalError.Error()
 }
 
+type verificationResult struct {
```
almostinf, line 34: Why do you need a new structure? There is `ErrrInvalidExpression` above and, as I understood, `verificationResult` can contain an error only if the expression is invalid, so `ErrrInvalidExpression` fits your needs quite well. But maybe I'm misunderstanding something, can you please explain it

expression/expression.go
```diff
@@ -77,6 +84,54 @@ func (triggerExpression TriggerExpression) Get(name string) (interface{}, error)
...
+// Validate returns error if triggers of type moira.ExpressionTrigger are badly formatted otherwise nil
+func (triggerExpression *TriggerExpression) Validate() error {
+	if triggerExpression.TriggerType != moira.ExpressionTrigger {
+		return fmt.Errorf("only advanced triggers should be validated")
+	}
+	if triggerExpression.Expression == nil {
+		return fmt.Errorf("trigger_type set to expression, but no expression provided")
+	}
+	expression := *triggerExpression.Expression
+	cacheKey := triggerExpression.validationCacheKey()
```
almostinf, line 96: Here I just want to clarify that I understood you correctly. You added this feature because the same expression can have different targets, right?

And further, if the expression has been checked:
- if there was an error, then you return it
- if there was no error, then why are you doing this: `validated = false`, if `validated` is not used anywhere else and why do you then delete this expression from the cache?

Also, if you chose the cache path, then maybe you should remove from the cache the expressions that change when the trigger is changed or deleted?

expression/expression.go
lordvidex: I will remove the caching of verificationResult in the next commit, therefore there will be no need for this struct.

expression/expression.go
lordvidex: Cache is no longer used but just to explain this scenario:

- if there is an error, it is returned
- if there is not but it has been checked `err.internalError` will also be nil and the function will also return with nil error.
- since we used a map containing interface{} to cache, we cannot guarantee the type, hence the check 
`err, ok := result.(verificationResult)`
and if the type cast was not successful, this means that the key is used to store value of another type in the cache, and this must be unexpected behavior, this is why I chose to delete it from the cache.

expression/expression.go
```diff
@@ -77,6 +84,54 @@ func (triggerExpression TriggerExpression) Get(name string) (interface{}, error)
 	}
 }
 
+// Validate returns error if triggers of type moira.ExpressionTrigger are badly formatted otherwise nil
+func (triggerExpression *TriggerExpression) Validate() error {
+	if triggerExpression.TriggerType != moira.ExpressionTrigger {
```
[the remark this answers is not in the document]
lordvidex: I think it will be better to use guard checks as there are many things we have already started guarding against inside the function. Like so:
```
	if triggerExpression.TriggerType != moira.ExpressionTrigger {
		return nil
	}
	if triggerExpression.Expression == nil || *triggerExpression.Expression == "" {
		return fmt.Errorf("trigger_type set to expression, but no expression provided")
	}
```

api/dto/triggers.go
```diff
@@ -195,7 +195,7 @@ func (trigger *Trigger) Bind(request *http.Request) error {
 
 	middleware.SetTimeSeriesNames(request, metricsDataNames)
 
-	if _, err := triggerExpression.Evaluate(); err != nil {
+	if err = triggerExpression.Validate(); err != nil {
```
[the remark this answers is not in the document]
lordvidex: I don't understand this. I think the Validate function should only do one thing. Since Evaluate returns a result we do not need, I think there is no need to call it.

expression/expression.go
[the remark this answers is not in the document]
almostinf: ok, no problem

api/dto/triggers.go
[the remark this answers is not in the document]
almostinf: I figured this way we maximize the chance of catching a bug if it's in runtime, but overall it's not that big of a deal

expression/expression.go
```diff
@@ -77,6 +79,41 @@ func (triggerExpression TriggerExpression) Get(name string) (interface{}, error)
 	}
 }
 
+// Validate returns error if triggers of type moira.ExpressionTrigger are badly formatted otherwise nil
+func (triggerExpression *TriggerExpression) Validate() error {
```
almostinf, line 83: Now looked at the front hanging when 400 is returned from `api/trigger/check`, this is fixed if you wrap the returned error in `ErrrInvalidExpression`, I have locally raised and tested this, so I suggest the following changes:
```go
// Validate returns error if triggers of type moira.ExpressionTrigger are badly formatted otherwise nil
func (triggerExpression *TriggerExpression) Validate() error {
	if triggerExpression.TriggerType != moira.ExpressionTrigger {
		return nil
	}
	if triggerExpression.Expression == nil || *triggerExpression.Expression == "" {
		return ErrInvalidExpression{internalError: fmt.Errorf("trigger_type set to expression, but no expression provided")}
	}
	expression := *triggerExpression.Expression
	env := map[string]interface{}{
		"ok":         moira.StateOK,
		"error":      moira.StateERROR,
		"warn":       moira.StateWARN,
		"warning":    moira.StateWARN,
		"nodata":     moira.StateNODATA,
		"t1":         triggerExpression.MainTargetValue,
		"prev_state": triggerExpression.PreviousState,
	}
	if triggerExpression.WarnValue != nil {
		env["warn_value"] = *triggerExpression.WarnValue
	}
	if triggerExpression.ErrorValue != nil {
		env["error_value"] = *triggerExpression.ErrorValue
	}
	for k, v := range triggerExpression.AdditionalTargetsValues {
		env[k] = v
	}
	if _, err := expr.Compile(
		strings.ToLower(expression),
		expr.Optimize(true),
		expr.Env(env),
	); err != nil {
		return ErrInvalidExpression{internalError: err}
	}
	return nil
}
```

lordvidex: What about errors like this one when checking for nil / empty Expression. Will that not affect the front as well? :
```		
return fmt.Errorf("trigger_type set to expression, but no expression provided")
```

almostinf: I just did it the way it was in `Evaluate`, there absolutely all errors are wrapped up in `ErrrInvalidExpression`

lordvidex: alright, fix coming in hot ))

perfomance_tests/expression/expression_test.go
Tetrergeru: :eyes:

lordvidex: I think this issue can be resolved already. @kissken ?

PR #896, opened 2023-08-17
fix(filter): Wildcard trigger compatibility with graphite-clickhouse

lordvidex:
- there is no need for the variable `allowMatchEmpty` because it breaks compatibility with `graphite-clickhouse`

Additional information
**BEFORE:** Triggers that contained tags like `tag1=~*` with wildcard matching was matching for triggers that didn't have tag1 at all 
**NOW:** When wildcard is used, at least the `tag1` key must be present in the trigger before it can be a valid match.

> View changes in tests to confirm new behaviour of matching tag patterns


filter/series_by_tag.go
```diff
@@ -63,7 +72,7 @@ func transformWildcardToRegexpInSeriesByTag(input string) (string, bool) {
 	}
 
 	if isTransformed {
-		result += "$"
+		result = fmt.Sprintf("^%s$", result)
```
msaf1980, line 75: May be simple strinng concat is better (Sprintf with dynamic format string has bad perfomance).

filter/series_by_tag.go
```diff
@@ -163,13 +170,26 @@ func createMatchingHandlerForOneTag(spec TagSpec) MatchingHandler {
...
+		if !strings.HasSuffix(value, "$") {
+			value += ".*"
+		}
+		matchRegex := regexp.MustCompile(value)
 		matchingHandlerCondition = func(value string) bool {
 			return matchRegex.MatchString(value)
 		}
 	case NotMatchOperator:
-		matchRegex := regexp.MustCompile("^" + spec.Value)
+		value := cleanAsterisks(spec.Value)
```
msaf1980, line 185: Why is needed ? Replace `*` to `.*` needed only if full string is `*`. Also may be add more tests for `*`

lordvidex: @msaf1980 I don't know if i missed any tests, I have tried the following special cases here: https://github.com/moira-alert/moira/pull/896/files#diff-4e74ec0972e1312934727b9a6b03635e81344c8d854205ea594290ad604bedc7

- empty
- single asterisks
- asterisk after an asterisk
- properly written asterisk should remain i.e. 
    - `.*` => `..*` ❌
    - `.*` => `.*` ✅
- when regexp has been anchored with `^` and `$`
- prefix and suffix asterisks
- mixed (correct and incorrectly written) asterisks

filter/series_by_tag.go
lordvidex: This is true, thanks. I've fixed in new commit.

filter/series_by_tag_pattern_index_test.go
```diff
@@ -225,7 +321,86 @@ func TestSeriesByTagPatternIndex(t *testing.T) {
...
+			},
+		}
+
+		testCases := []struct {
+			Name            string
+			Labels          map[string]string
+			MatchedPatterns []string
+		}{
+			{
+				"something",
```
almostinf, line 377: maybe change the name to something more revealing? :)

filter/series_by_tag.go
msaf1980: In graphite-clickhouse seriesByTags `*` replaced to `.*` only in wildcards (it's not regular expression, but translated to regexp for check). One case whan `*` translated to '.*` in regexp, is a bad regexp `*`.

msaf1980: Also, add tests for regexp like `a*b` and `a.*b` for check why replacement `*` to `.*` is a bad thing.

lordvidex: Done

filter/series_by_tag.go
```diff
@@ -179,14 +214,14 @@ func createMatchingHandlerForOneTag(spec TagSpec) MatchingHandler {
 		}
 	}
 
-	matchEmpty := matchingHandlerCondition("")
+	// matchEmpty := matchingHandlerCondition("")
```
kissken, line 217: is it ok?

lordvidex: I want to remove it, but I need @msaf1980 to check graphite-clickhouse and give a review about some errors that might depend on this.

filter/series_by_tag.go
```diff
@@ -63,7 +69,7 @@ func transformWildcardToRegexpInSeriesByTag(input string) (string, bool) {
 	}
 
 	if isTransformed {
-		result += "$"
+		result = "^" + result + "$"
```
Tetrergeru, line 72:
```suggestion
		result = fmt.Sprintf("^%s$", result)"
```

Tetrergeru:
We had to split this PR into two parts, to deploy them seprately: #924 and #923, so I am going to close this PR

@lordvidex, I mentioned you in these PR descriptions, thank you very much for your contribution!

lordvidex:
You're welcome!

PR #907, opened 2023-09-06
feature(api): add refiller index goroutine for search index

almostinf:
# Refiller goroutine

Repopulates the search index thereby removing memory leaks that appeared when filling, updating and searching the index. Also removed memory leaks in tests and refactoring

**TODO**: add tests for refiller


cmd/api/main.go
```diff
@@ -84,6 +84,7 @@ func main() {
 	databaseSettings := applicationConfig.Redis.GetSettings()
 	notificationHistorySettings := applicationConfig.NotificationHistory.GetSettings()
 	database := redis.NewDatabase(logger, databaseSettings, notificationHistorySettings, redis.API)
+	database.Flush()
```
kissken, line 87: ой

PR #909, opened 2023-09-11
feat Add prometheus retries

Tetrergeru:


metric_source/prometheus/fetch.go
```diff
@@ -15,7 +15,28 @@ import (
...
+		if err == nil {
+			return res, nil
+		}
+
+		prometheus.logger.Warning().
+			Error(err).
+			Int("retries left", prometheus.config.Retries-i-1).
+			Msg("Failed to fetch prometheus target")
+
+		time.Sleep(prometheus.config.RetryTimeout)
```
kissken, line 32: why should we sleep if it is last available attempt?

Tetrergeru: Fixed

kissken (approved):

kissken (approved):

PR #915, opened 2023-09-14
add api for get unused triggers

kissken:
for getting triggers without subs

### DIFF
--- a/api/controller/triggers.go
+++ b/api/controller/triggers.go
@@ -113,7 +113,7 @@ func SearchTriggers(database moira.Database, searcher moira.Searcher, page int64
 		searchResults = searchResults[from:to]
 	}
 
-	triggerIDs := make([]string, 0)
+	triggerIDs := make([]string, 0, len(searchResults))
 	for _, searchResult := range searchResults {
 		triggerIDs = append(triggerIDs, searchResult.ObjectID)
 	}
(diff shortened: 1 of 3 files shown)


api/controller/triggers.go
```diff
@@ -113,7 +113,7 @@ func SearchTriggers(database moira.Database, searcher moira.Searcher, page int64
 		searchResults = searchResults[from:to]
 	}
 
-	var triggerIDs []string //nolint
+	triggerIDs := make([]string, 0)
```
almostinf, line 116: Here it would be better to set the capacity

api/controller/triggers.go
```diff
@@ -151,24 +151,57 @@ func SearchTriggers(database moira.Database, searcher moira.Searcher, page int64
...
+
+	return triggersList, nil
+}
+
+func getTriggerChecks(database moira.Database, triggerIDs []string) ([]moira.TriggerCheck, error) {
+	triggerChecks, err := database.GetTriggerChecks(triggerIDs)
+	if err != nil {
+		return nil, err
+	}
+	list := make([]moira.TriggerCheck, 0)
```
almostinf, line 192: Here, too, it's like it's better to add capacity

Tetrergeru (dismissed):

api/controller/triggers.go
kissken: fixed

api/controller/triggers.go
kissken: fixed

almostinf (approved):

Tetrergeru (approved):

PR #946, opened 2023-10-24
feat(notifier, api): fix sending notifications on muted and deleted triggers

almostinf:
# Fix sending notifications on muted and deleted triggers and muted metrics

This PR solves the problem of unwanted notifications being sent on triggers or metrics to Maintenance or removed triggers

To do this, for each notification we collect a `checkData` array (optimized for the case where there are many notifications with the same timestamp). If there is no trigger, the corresponding `checkData` = nil, whether a trigger or a metric is on Maintenance can be found by the Maintenance field

For optimization purposes, all notifications are divided into delayed and not delayed notifications and all checks needed for validation are performed on delayed notifications. Also all validations are done in a single transaction to avoid data race

Notification types have also been added:
- `valid` - correct notifications
- `removed` - incorrect notifications (for example, with a removed trigger).
- `ignored` - notifications whose trigger or metric is on Maintenance

Divided 1 big PR into 3 parts, here are the rest:
- #943
- #944

PR #952, opened 2023-10-31
fix(notifier): fix fetching notifications by batch

almostinf:
# Fix fetching notifications by batch

After the merge PR #946 it will be broken to receive notifications with a limit, this PR solves that problem

I replaced the `IgnoredNotification` state with `ResendNotification` so that all notifications whose trigger or metric is on Maintenance are resaved in the database, this is to avoid creating a large queue of such notifications with an unupdated Timestamp in the database because they would interfere with fetching notifications with a limit. Now all such notifications will be deleted, and their copies with updated Timestamp on `resave_time` will be stored in the database


database/redis/notification.go
```diff
@@ -146,8 +139,16 @@ func (connector *DbConnector) GetDelayedTimeInSeconds() int64 {
 	return int64(connector.notification.DelayedTime.Seconds())
 }
 
+// GetResaveTimeInSeconds returns the time to reschedule notifications to the future in seconds
+func (connector *DbConnector) GetResaveTimeInSeconds() int64 {
```
Tetrergeru, line 143: А мы её снаружи испольщуем? Нельзя заприватить?

database/redis/notification.go
```diff
@@ -511,7 +514,27 @@ func (connector *DbConnector) AddNotifications(notifications []*moira.ScheduledN
...
 
 	if err != nil {
-		return fmt.Errorf("failed to EXEC: %s", err.Error())
+		return fmt.Errorf("failed to EXEC: %w", err)
+	}
+
+	return nil
+}
+
+// saveNotifications save notifications with the passed pipe and context
```
Tetrergeru, line 523: Зачем годок к приватной функции?

database/redis/notification_test.go
```diff
@@ -1001,6 +1014,9 @@ func TestFetchNotificationsDo(t *testing.T) {
 			dataBase.SetTriggerCheckMaintenance("test2", map[string]int64{"test2": time.Now().Add(time.Hour).Unix()}, nil, "test", 100) //nolint
 			defer dataBase.SetTriggerCheckMaintenance("test2", map[string]int64{"test2": 0}, nil, "test", 100)                          //nolint
 
+			updatedNotificationNew3 := notificationNew3
+			updatedNotificationNew3.Timestamp += dataBase.GetResaveTimeInSeconds()
```
Tetrergeru, line 1018: Может сразу переименуем `dataBase` -> `database`? Мне очень глаза режет :sweat_smile:

database/redis/notification.go
almostinf: Заприватил

database/redis/notification.go
almostinf: Убрал

database/redis/notification_test.go
almostinf: Переименовал)

PR #957, opened 2023-11-07
feat(api): added full trigger validation in api

Tetrergeru:
Previously triggerexpression were evaluated via lazy evaluation, this caoused some problems for validation e.g. branches that were not reached durina evaluation were not typechecked.

Now we change package for expression evaluation and all expressions will be checked stricktly


AleksandrMatsko (approved):
Логических отступов бы наставить, а так аппрув

kissken (approved):

kissken (approved):

PR #970, opened 2023-11-23
fix(api): fix incorrect display of deleted but kept metrics

almostinf:
# Fix incorrect display of deleted but kept metrics

Fixing a bug that appeared after this #916 when metrics that are deleted but kept are not displayed on `/api/trigger/{id}/state`, but are displayed on `/api/trigger`, `/api/trigger/search` and `/api/trigger/unused`. Also added tests on it

This logic must be in the controller layer, as changes to the database will break the logic of the checker, which expects it to get all metrics


api/controller/triggers.go
```diff
@@ -191,6 +196,9 @@ func getTriggerChecks(database moira.Database, triggerIDs []string) ([]moira.Tri
 	list := make([]moira.TriggerCheck, 0, len(triggerChecks))
 	for _, triggerCheck := range triggerChecks {
 		if triggerCheck != nil {
+			if len(triggerCheck.LastCheck.Metrics) != 0 {
```
kissken, line 199: не лучше перенести это условие вовнутрь функции, чтобы она сразу возвращала нужное?

kissken (approved):

PR #971, opened 2023-12-05
feat(senders): add templating body in webhook 

almostinf:
# Add ability to make template body in webhook

Added the ability to template the request body in the webhook in the config

If you do not specify the `body` field, the request body will remain the same, and if you specify it, we will take the templated one from the config


datatypes.go
```diff
@@ -197,6 +201,15 @@ type ContactData struct {
 	Team  string `json:"team"`
 }
 
+func (contact *ContactData) ToTemplatingContactInfo() templating.ContactInfo {
```
kissken, line 204: заприватить?

PR #994, opened 2024-02-19
feat(notifier): add ability to scale senders

### LINKED ISSUE #821: Add support of multiple senders with type: mail
# FEATURE
Notifier supports only one sender with type mail, but sometimes there is such situation when you need to send alerting emails from different addresses/to different smtp connections/with different templates/etc
## Summary

Main idea is to give the ability to set up multiple email senders, each one with its own settings.It seems to me that it would be convenient to do it by the same mechanism like it done with webhooks -  with differentiation by field ```name```.

### Details

<details>
I've made some experiments with adding multiple email senders. It looks like that connection that goes lower in notifier.yml just overrides upper connection settings.
</details>

almostinf:
# Add ability to scale senders

Now each sender is identified by a unique `contact_type` from sender settings, this allowed to painlessly solve problems with multiple senders of the same type

Also unified configs for all senders by removing the required `name` field in webhook and script configurations

Closes #821


notifier/notifier_test.go
```diff
@@ -19,17 +19,31 @@ import (
...
+		Senders: []map[string]interface{}{
+			{
+				"type":       "test_type",
+				"contact_id": "test_contact_id",
+			},
+		},
+	}
 )
 
 var (
```
kissken, line 40: может в одну var все схлопуть?

almostinf: Готово

kissken (approved):

notifier/notifier_test.go
```diff
@@ -19,17 +19,29 @@ import (
...
-	plots    [][]byte
-	shutdown = make(chan struct{})
-)
+	plots          [][]byte
+	shutdown       = make(chan struct{})
+	location, _    = time.LoadLocation("UTC")
```
Tetrergeru, line 24: Может всё, что нам не надо менять сделаем константами?

Tetrergeru (approved):

notifier/notifier_test.go
almostinf: Сделал

PR #1069, opened 2024-08-06
fix(notifier): fixed sending of notifications with schedule between different days

almostinf:
# Fixed sending of notifications with schedule between different days

The problem occurred when sending notifications in a subscription with a schedule of 11:30 PM - 06:00 PM in which the end was earlier than the beginning, i.e. it was supposed to go to the next day

In the test cases there was a similar example with the schedule from 02:00 AM - 00:00 AM earlier, but in this case everything happens **within one day**, which did not allow to detect the bug

In this PR I have fixed this incorrect behaviour by adding additional handling for the case where the end of the schedule is earlier than the beginning. Let's look at an example:
- We have a subscription with a schedule from 11:30 PM - 06:00 PM
- A metric came to us that generated an alert at 02:00 PM
- Before this PR we would only increment the 06:00 PM to the next day, not taking into account that the 11:30 PM subscription start is actually supposed to refer to the previous day, not the current day, accordingly in my new PR we would be pushing the start back a day. The code checks that the number of hours that have passed since the start of the new day is less than the subscription start time and the subscription end time, in which case we need to move the start of the subscription back by a day
diff_days

Let's consider another example that worked correctly:
- We have a subscription with a schedule from 11:30 PM - 06:00 PM
- A metric came to us and generated an alert at 11:45 PM
- In this case, the logic before this PR was correct, as the start of the subscription is within the **current day**, and the end is just **on another day**, so the end needs to be incremented one day forward
one_day

### DIFF
--- a/notifier/scheduler.go
+++ b/notifier/scheduler.go
@@ -162,16 +162,24 @@ func calculateNextDelivery(schedule *moira.ScheduleData, nextTime time.Time) (ti
 	if len(schedule.Days) == 0 {
 		return nextTime, nil
 	}
+
 	beginOffset := time.Duration(schedule.StartOffset) * time.Minute
 	endOffset := time.Duration(schedule.EndOffset) * time.Minute
-	if schedule.EndOffset < schedule.StartOffset {
-		endOffset += time.Hour * 24
-	}
 
 	tzOffset := time.Duration(schedule.TimezoneOffset) * time.Minute
 	localNextTime := nextTime.Add(-tzOffset).Truncate(time.Minute)
 	localNextTimeDay := localNextTime.Truncate(24 * time.Hour) //nolint
 	localNextWeekday := int(localNextTimeDay.Weekday()+6) % 7  //nolint
+
+	if schedule.EndOffset < schedule.StartOffset {
+		// The condition can only be fulfilled if the begin offset should be on the past day and not on the current day.
+		// In other variants end offset must be on the next day
+		if localNextTime.Sub(localNextTimeDay) < beginOffset && localNextTime.Sub(localNextTimeDay) < endOffset {
+			beginOffset -= time.Hour * 24
+		} else {
+			endOffset += time.Hour * 24
+		}
+	}
 
 	if schedule.Days[localNextWeekday].Enabled &&
 		(localNextTime.Equal(localNextTimeDay.Add(beginOffset)) || localNextTime.After(localNextTimeDay.Add(beginOffset))) &&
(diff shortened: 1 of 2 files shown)


notifier/scheduler.go
```diff
@@ -162,17 +162,25 @@ func calculateNextDelivery(schedule *moira.ScheduleData, nextTime time.Time) (ti
...
 	tzOffset := time.Duration(schedule.TimezoneOffset) * time.Minute
 	localNextTime := nextTime.Add(-tzOffset).Truncate(time.Minute)
 	localNextTimeDay := localNextTime.Truncate(24 * time.Hour) //nolint
 	localNextWeekday := int(localNextTimeDay.Weekday()+6) % 7  //nolint
 
+	if schedule.EndOffset < schedule.StartOffset {
+		// The condition can only be fulfilled if the begin offset should be on the past day and not on the current day.
+		// In other variants end offset must be on the next day
+		if localNextTime.Sub(localNextTimeDay) < beginOffset && localNextTime.Sub(localNextTimeDay) < endOffset {
+			beginOffset -= time.Hour * 24
```
kissken, line 178: мб утащим в вар `time.Hour * 24`?

Tetrergeru: Мне кажется, что оно и так нормально читается. 24 часа = день

notifier/scheduler.go
```diff
@@ -162,17 +162,25 @@ func calculateNextDelivery(schedule *moira.ScheduleData, nextTime time.Time) (ti
...
 
 	tzOffset := time.Duration(schedule.TimezoneOffset) * time.Minute
 	localNextTime := nextTime.Add(-tzOffset).Truncate(time.Minute)
 	localNextTimeDay := localNextTime.Truncate(24 * time.Hour) //nolint
 	localNextWeekday := int(localNextTimeDay.Weekday()+6) % 7  //nolint
 
+	if schedule.EndOffset < schedule.StartOffset {
+		// The condition can only be fulfilled if the begin offset should be on the past day and not on the current day.
+		// In other variants end offset must be on the next day
+		if localNextTime.Sub(localNextTimeDay) < beginOffset && localNextTime.Sub(localNextTimeDay) < endOffset {
```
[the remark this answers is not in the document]
almostinf: Вынес в переменную

notifier/scheduler.go
almostinf: Мне кажется, да, алиас излишен

Tetrergeru (approved):

AleksandrMatsko (approved):

PR #1073, opened 2024-08-21
feat: trigger description to telegram

### LINKED ISSUE #353: Markdown descriptions in senders

AleksandrMatsko:
# Add formatted trigger description to alerts for telegram 

Change the way of cutting description in `HighlightSyntaxFormatter`. 
Description for telegram is formatted by using html. If it's size is too long then it will be cut to feat telegram limits. 

Closes #353


AleksandrMatsko:
After some discussion other way of adding trigger description will be used

PR #1113, opened 2024-10-09
feat(api): print trigger id when it already exists

HeavyPunk:
# PR Summary

Added a trigger id to api response on creating trigger (PUT /api/trigger/)

## Additional information

None

## Closes/Relates #issue

None


kissken (approved):
👍🏻

PR #1122, opened 2024-11-12
refactor: team validation

AleksandrMatsko:
# PR Summary

Before this PR it was possible to create teams with same names.

This PR forbids create/update of team with name which is alredy belongs to other team (no matter case, so names *"Team name"* and *"Team Name"* are the **same** then checking name existence).

Also:
- Refactor team limits. Move limits to `LimitsConfig`.
- Add tests for teams in dto package.

Note: Database schema is changed a bit. Added separate key (`"moira-teams-by-names"`) to store team IDs by team names. Than saving/deleting team this key is also updated. 

## updating from 2.13
Use following command to update to new version.
```bash
cli --update --from-version '2.13'
```
This command will do the following:
1. Get all teams from DB.
2. Group teams with same names.
3. For teams with same names change name (example: ["name", "name", "name"] -> ["name", "name1", "name2"]).
4. Update teams with name changed.
5. Save pairs teamName:team.ID to `"moira-teams-by-names"` redis hash.

## downgrading to 2.13
Use following command to downgrade to previous version.
```bash
cli --downgrade --to-version '2.13'
```
This command will delete `"moira-teams-by-names"` redis hash.


AleksandrMatsko:
Split into #1123 and #1124

